PR #1, opened 2017-12-16
add an example yamlfile to readme

hallyn:
Might not be where you want it, just an idea :)


tych0:
Thanks!

PR #7, opened 2018-03-02
add oci basetype

hallyn:
Format is:

  type: oci
  url: ocilayoutdir:imagename

### DIFF
--- a/base.go
+++ b/base.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"strings"
 
 	"github.com/openSUSE/umoci"
 )
@@ -26,7 +27,7 @@ func GetBaseLayer(o BaseLayerOpts) error {
 	case TarType:
 		return getTar(o)
 	case OCIType:
-		return fmt.Errorf("not implemented")
+		return getOci(o)
 	case DockerType:
 		return getDocker(o)
 	case ScratchType:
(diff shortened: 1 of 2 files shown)


base.go
```diff
@@ -36,6 +37,75 @@ func GetBaseLayer(o BaseLayerOpts) error {
...
+		tag = res[1]
+	}
+
+	cacheDir := path.Join(o.Config.StackerDir, "layer-bases", tag)
+	if err := os.MkdirAll(cacheDir, 0755); err != nil {
+		return err
+	}
+
+	cmd := exec.Command(
+		"skopeo",
```
tych0: I think you need an `--insecure-policy` here to keep everyone from having to add /etc/container/policy.json to their systems.

base.go
```diff
@@ -36,6 +37,75 @@ func GetBaseLayer(o BaseLayerOpts) error {
...
+	)
+
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	err = cmd.Run()
+	if err != nil {
+		return fmt.Errorf("skopeo copy: %s", err)
+	}
+
+	// We just copied it to the cache, now let's copy that over to our image.
```
tych0: Any chance we can abstract this since it's copy pasted from something else?

hallyn: Yeah I'm going to close this PR - I think the abstraction needs to be fixed up.  The 'url' is not generic enough.  We want a type struct that has a generic 'source' struct (which can be a dockerhub url or an oci layout directory), an image name, and maybe optional additional tag.

PR #11, opened 2018-03-29
Add a Dockerfile for stacker.

kprav33n:
Add a Dockerfile to build docker image. This will help folks who want to develop on environments where they don't want to run vagrant. For instance, a developer on macOS can use xhyve/Docker instead of using VirtualBox, or VMWare Fusion.


tych0:
Thanks!

PR #12, opened 2018-04-17
fix docker path

hallyn:
commit 4374239cca broke some builds by trying to skopeo copy
images to, say, "oci:oci:anuvu/anuvu-dev".  That's because
it dropped the 'path.Base(anuvu/anuvu-dev)'.  Re-add that.


tych0:
I knew I should have kept this...

PR #13, opened 2018-06-05
use stat to determine bind mount type

hallyn:
If we bind mount a symlink to a directory, we want
the target directory bind mounted.


tych0:
Thanks!

PR #18, opened 2018-10-02
stacker: ensure that the dest dir exists in importFile()

pcmoore:
Signed-off-by: Paul Moore <paul@paul-moore.com>


tych0:
Thanks! Derp moment for me, it seems :)

PR #20, opened 2018-11-20
Fixed directory import

CajuM:
When a parent directory is imported after it's children, it is removed and recreated.
When the "." directory gets imported it's destination path is incorrect.
Now the import logic only calls cp -a on files.


tych0:
Awesome, thanks!

PR #22, opened 2019-01-10
doc: update to reference golang 1.10 and libcap-dev dependency.

smoser:
The reference to golang 1.9 is incorrect now as things have moved on.
libcap-dev is needed for sys/capability.h


tych0:
Cool, thanks!

PR #23, opened 2019-01-11
Mark dependencies for 'make check'.

smoser:
'make check' uses bats package for unit tests and jq inside the unit
tests for verification.


tych0:
Thanks!

PR #24, opened 2019-01-11
Fix make check failure due to order of arguments to sudo.

smoser:
It should be:
 sudo ENV=VAL command
not
 sudo command ENV=val


tych0:
Thanks!

PR #33, opened 2019-05-15
Pr fixes

CajuM:
This PR introduces two changes:
* change the .working container name to _working as valid dns names can't begin with a `.` this messes up some test suites
* add /dev/shm support, as it exists on standard Linux hosts


tych0:
Thanks!

PR #35, opened 2019-05-31
Building multiple stacker recipes in same execution

andaaron:
1. Changes in this PR
- Update the stacker recipe to add a 'stacker_config' key. The 'stacker_config' contains build configuration for all the layers in the stacker recipe. For now the only key in 'stacker_config' is 'include'. 'include' contains paths to other stacker recipes which contain layers which should be built before the current stacker recipe.
- Modified the 'stacker build' command to also build the layers mentioned in the 'include' section of the stacker recipe (and other includes found in those recipes, recursively). 
This behavior can be disabled with a '--no-includes' flag.
- Other changes which are prerequisites for the above.

2. If this PR is OK, we can move forward and:
- Add a separate command to search for stacker recipes in the file system using a regex, and arrange them in order according to dependencies, and build them. Or we can add another set of flags to indicate a pattern / search folder to the build command itself, which would be mutually exclusive with '-f'. 
- Come up with a mechanism for caching layers in dockerhub.

### DIFF
--- a/api.go
+++ b/api.go
@@ -377,7 +377,7 @@ func NewStackerfile(stackerfile string, substitutions []string) (*Stackerfile, e
 			return nil, err
 		}
 
-		// This file is on the disk, get it's current directory
+		// This file is on the disk, use its parent directory
 		sf.referenceDirectory = filepath.Dir(sf.path)
 
 	} else {
@@ -396,7 +396,7 @@ func NewStackerfile(stackerfile string, substitutions []string) (*Stackerfile, e
 			return nil, err
 		}
 
-		// There's no need to update the working directory of the stackerfile
+		// There's no need to update the reference directory of the stackerfile
 		// Continue to use the working directory
 	}
 
@@ -493,49 +493,6 @@ func NewStackerfile(stackerfile string, substitutions []string) (*Stackerfile, e
 	return &sf, err
 }
 
-// NewStackerFiles reads multiple Stackerfiles from a list of paths and applies substitutions
-// It adds the Stackerfiles mentioned in the prerequisite paths to the results
-func NewStackerFiles(paths []string, substituteVars []string) (map[string]*Stackerfile, error) {
-	sfm := make(map[string]*Stackerfile, len(paths))
-
-	// Iterate over list of paths to stackerfiles
-	for _, path := range paths {
-		fmt.Printf("initializing stacker recipe: %s\n", path)
-
-		// Read this stackerfile
-		sf, err := NewStackerfile(path, substituteVars)
-		if err != nil {
-			return nil, err
-		}
-
-		// Add using absolute path to make sure the entries are unique
-		absPath, err := filepath.Abs(path)
-		if err != nil {
-			return nil, err
-		}
-		if _, ok := sfm[absPath]; ok != true {
-			sfm[absPath] = sf
-		}
-
-		// Determine correct path of prerequisites
-		prerequisites, err := sf.Prerequisites()
-		if err != nil {
-			return nil, err
-		}
-
-		// Need to also add stackerfile dependencies of this stackerfile to the map of stackerfiles
-		depStackerFiles, err := NewStackerFiles(prerequisites, substituteVars)
-		if err != nil {
-			return nil, err
-		}
-		for depPath, depStackerFile := range depStackerFiles {
-			sfm[depPath] = depStackerFile
-		}
-	}
-
-	return sfm, nil
-}
-
 // DependencyOrder provides the list of layer names from a stackerfile
 // the current order to be built, note this method does not reorder the layers,
 // but it does validate they are specified in an order which makes sense
@@ -630,4 +587,62 @@ func (sf *Stackerfile) Prerequisites() ([]string, error) {
 		}
 	}
 	return prerequisitePaths, nil
+}
+
+// Logic for working with multiple StackerFiles
+type StackerFiles map[string]*Stackerfile
+
+// NewStackerFiles reads multiple Stackerfiles from a list of paths and applies substitutions
+// It adds the Stackerfiles mentioned in the prerequisite paths to the results
+func NewStackerFiles(paths []string, substituteVars []string) (StackerFiles, error) {
+	sfm := make(map[string]*Stackerfile, len(paths))
+
+	// Iterate over list of paths to stackerfiles
+	for _, path := range paths {
+		fmt.Printf("initializing stacker recipe: %s\n", path)
+
+		// Read this stackerfile
+		sf, err := NewStackerfile(path, substituteVars)
+		if err != nil {
+			return nil, err
+		}
+
+		// Add using absolute path to make sure the entries are unique
+		absPath, err := filepath.Abs(path)
+		if err != nil {
+			return nil, err
+		}
+		if _, ok := sfm[absPath]; ok != true {
+			sfm[absPath] = sf
+		}
+
+		// Determine correct path of prerequisites
+		prerequisites, err := sf.Prerequisites()
+		if err != nil {
+			return nil, err
+		}
+
+		// Need to also add stackerfile dependencies of this stackerfile to the map of stackerfiles
+		depStackerFiles, err := NewStackerFiles(prerequisites, substituteVars)
+		if err != nil {
+			return nil, err
+		}
+		for depPath, depStackerFile := range depStackerFiles {
+			sfm[depPath] = depStackerFile
+		}
+	}
+
+	return sfm, nil
+}
+
+// LookupLayerDefinition searches for the Layer entry within the Stackerfiles
+func (sfm StackerFiles) LookupLayerDefinition(name string) (*Layer, bool) {
+	// Search for the layer in all of the stackerfiles
+	for _, sf := range sfm {
+		l, found := sf.Get(name)
+		if found {
+			return l, true
+		}
+	}
+	return nil, false
 }
--- a/api.go
+++ b/api.go
@@ -6,7 +6,9 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"os"
 	"path"
+	"path/filepath"
 	"reflect"
 	"regexp"
 	"strings"
@@ -29,6 +31,10 @@ type StackerConfig struct {
 	RootFSDir  string `yaml:"rootfs_dir"`
 }
 
+type BuildConfig struct {
+	Prerequisites []string `yaml:"prerequisites"`
+}
+
 type Stackerfile struct {
 	// AfterSubstitutions is the contents of the stacker file after
 	// substitutions (i.e., the content that is actually used by stacker).
@@ -39,6 +45,15 @@ type Stackerfile struct {
 
 	// fileOrder is the order of elements as they appear in the stackerfile.
 	fileOrder []string
+
+	// configuration specific for this specific build
+	buildConfig *BuildConfig
+
+	// path to stackerfile
+	path string
+
+	// directory relative to which the stackerfile content is referenced
+	referenceDirectory string
 }
 
 func (sf *Stackerfile) Get(name string) (*Layer, bool) {
@@ -98,19 +113,20 @@ func (is *ImageSource) ParseTag() (string, error) {
 }
 
 type Layer struct {
-	From        *ImageSource      `yaml:"from"`
-	Import      interface{}       `yaml:"import"`
-	Run         interface{}       `yaml:"run"`
-	Cmd         interface{}       `yaml:"cmd"`
-	Entrypoint  interface{}       `yaml:"entrypoint"`
-	FullCommand interface{}       `yaml:"full_command"`
-	Environment map[string]string `yaml:"environment"`
-	Volumes     []string          `yaml:"volumes"`
-	Labels      map[string]string `yaml:"labels"`
-	WorkingDir  string            `yaml:"working_dir"`
-	BuildOnly   bool              `yaml:"build_only"`
-	Binds       interface{}       `yaml:"binds"`
-	Apply       []string          `yaml:"apply"`
+	From               *ImageSource      `yaml:"from"`
+	Import             interface{}       `yaml:"import"`
+	Run                interface{}       `yaml:"run"`
+	Cmd                interface{}       `yaml:"cmd"`
+	Entrypoint         interface{}       `yaml:"entrypoint"`
+	FullCommand        interface{}       `yaml:"full_command"`
+	Environment        map[string]string `yaml:"environment"`
+	Volumes            []string          `yaml:"volumes"`
+	Labels             map[string]string `yaml:"labels"`
+	WorkingDir         string            `yaml:"working_dir"`
+	BuildOnly          bool              `yaml:"build_only"`
+	Binds              interface{}       `yaml:"binds"`
+	Apply              []string          `yaml:"apply"`
+	referenceDirectory string            // Location of the directory where the layer is defined
 }
 
 func (l *Layer) ParseCmd() ([]string, error) {
@@ -132,21 +148,77 @@ func (l *Layer) ParseFullCommand() ([]string, error) {
 }
 
 func (l *Layer) ParseImport() ([]string, error) {
-	return l.getStringOrStringSlice(l.Import, func(s string) ([]string, error) {
+	rawImports, err := l.getStringOrStringSlice(l.Import, func(s string) ([]string, error) {
 		return strings.Split(s, "\n"), nil
 	})
+	if err != nil {
+		return nil, err
+	}
+
+	var absImports []string
+	for _, rawImport := range rawImports {
+		absImport, err := l.getAbsPath(rawImport)
+		if err != nil {
+			return nil, err
+		}
+		absImports = append(absImports, absImport)
+	}
+	return absImports, nil
 }
 
-func (l *Layer) ParseBinds() ([]string, error) {
-	return l.getStringOrStringSlice(l.Binds, func(s string) ([]string, error) {
+func (l *Layer) ParseBinds() (map[string]string, error) {
+	rawBinds, err := l.getStringOrStringSlice(l.Binds, func(s string) ([]string, error) {
 		return []string{s}, nil
 	})
+	if err != nil {
+		return nil, err
+	}
+
+	absBinds := make(map[string]string, len(rawBinds))
+	for _, bind := range rawBinds {
+		parts := strings.Split(bind, "->")
+		if len(parts) != 1 && len(parts) != 2 {
+			return nil, fmt.Errorf("invalid bind mount %s", bind)
+		}
+
+		source := strings.TrimSpace(parts[0])
+		target := source
+
+		absSource, err := l.getAbsPath(source)
+		if err != nil {
+			return nil, err
+		}
+
+		if len(parts) == 2 {
+			target = strings.TrimSpace(parts[1])
+		}
+
+		absBinds[absSource] = target
+	}
+
+	return absBinds, nil
+
 }
 
 func (l *Layer) ParseRun() ([]string, error) {
 	return l.getStringOrStringSlice(l.Run, func(s string) ([]string, error) {
 		return []string{s}, nil
 	})
+}
+
+func (l *Layer) getAbsPath(path string) (string, error) {
+	parsedPath, err := url.Parse(path)
+	if err != nil {
+		return "", err
+	}
+
+	if parsedPath.Scheme != "" || filepath.IsAbs(path) {
+		// Path is already absolute or is an URL, return it
+		return path, nil
+	} else {
+		// If path is relative we need to add it to the directory where this layer is found
+		return filepath.Abs(filepath.Join(l.referenceDirectory, path))
+	}
 }
 
 func (l *Layer) getStringOrStringSlice(iface interface{}, xform func(string) ([]string, error)) ([]string, error) {
@@ -275,7 +347,17 @@ func substitute(content string, substitutions []string) (string, error) {
 // explicitly not a map, because the substitutions are performed one at a time
 // in the order that they are given.
 func NewStackerfile(stackerfile string, substitutions []string) (*Stackerfile, error) {
+	var err error
+
 	sf := Stackerfile{}
+	sf.path = stackerfile
+
+	// Use working directory as default folder relative to which files
+	// in the stacker yaml will be searched for
+	sf.referenceDirectory, err = os.Getwd()
+	if err != nil {
+		return nil, err
+	}
 
 	url, err := url.Parse(stackerfile)
 	if err != nil {
@@ -285,6 +367,19 @@ func NewStackerfile(stackerfile string, substitutions []string) (*Stackerfile, e
 	var raw []byte
 	if url.Scheme == "" {
 		raw, err = ioutil.ReadFile(stackerfile)
+		if err != nil {
+			return nil, err
+		}
+
+		// Make sure we use the absolute path to the Stackerfile
+		sf.path, err = filepath.Abs(stackerfile)
+		if err != nil {
+			return nil, err
+		}
+
+		// This file is on the disk, get it's current directory
+		sf.referenceDirectory = filepath.Dir(sf.path)
+
 	} else {
 		resp, err := http.Get(stackerfile)
 		if err != nil {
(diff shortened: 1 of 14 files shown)


lib/dag.go
```diff
@@ -0,0 +1,177 @@
+package lib
+
+import (
+	"fmt"
+
+	"github.com/twmb/algoimpl/go/graph"
+)
+
+// This file is a copy from cube/gobi to avoid deps on cube/gobi
```
tych0: cube/gobi are not public, so this comment doesn't make a lot of sense.

build.go
```diff
@@ -680,10 +702,77 @@ func Build(opts *BuildArgs) error {
...
+	dag, err := NewStackerFilesDAG(stackerFiles)
+	if err != nil {
+		return err
+	}
+
+	// In case we don't want to build the images found in includes, filter them out
+	sortedPaths := dag.Sort()
+	var finalPaths []string
+	if !opts.NoIncludes {
+		finalPaths = sortedPaths
```
tych0: I think here, if len(sortedPaths) != len(paths), we should return errors.Errorf("includes specifically disallowed"), right?

build.go
```diff
@@ -680,10 +702,77 @@ func Build(opts *BuildArgs) error {
...
+				}
+				if sp == ap {
+					finalPaths = append(finalPaths, sp)
+				}
+			}
+		}
+	}
+
+	// Show the serial build order
+	fmt.Printf("stacker build order:\n")
```
tych0: @mikemccracken suggested that we have a separate command to render just the build order. Since we have this logic already, seems like that should be fairly easy?

api.go
```diff
@@ -32,7 +32,7 @@ type StackerConfig struct {
 }
 
 type BuildConfig struct {
-	Include    []string `yaml:"include"`
```
tych0: Can you squash this commit into the relevant other commits?

tych0:
One other thing we talked about offline but I'll record here for posterity is the use of `Chdir()`, we want to switch everything to just compute the absolute paths in `NewStackerfile()` so we don't have to `Chdir()` around everywhere.

build.go
andaaron: Actually that's not the intention.

The reason I added this option is in case someone knows he build the layers in the includes separately, and only wants to build the ones defined in the actual stackerfile sent as parameter.

If we add len() check he won't be able to do that, he'd have to remove/comment the includes altogether.
Do we want this?

tych0: What do you mean by "built them separately"? Surely we want to rebuild if things have changed according to the cache, right? Or maybe I don't understand.

andaaron: So let's say we have:
- 3 layers, A, B and C. A is base for B, B is base for C.
- A is on one repo. B and C are on another repo. 
So the include will be only in the stacker.yaml of image C, and mention image B.
- The stacker yaml of B has a docker url to image A in 'from'. 
The stacker yaml of C has a docker url to image B in 'from'.

In this scenario 
1. If I run the build for C with the default options it would:
- download C, if not already present in the cache
- run the build for B
- run the build for C, but because of the 'include' directive, the build would ignore the docker url to B, and use the one in the oci layout instead, or ignore the url in the stacker yaml and use the one provided for caching with a git-<commit> tag.
(this last part if not implemented right now, but I was thinking about it)

2. If I run the build for C with the '--no-includes' option it would:
- download B, if not already present in the cache
- run the build for C, ignoring the include directive, and basically behave as if the stacker.yaml did not have it.

So for the same updated stacker.yamls containing includes, 1. would enable us to use the git-commit caching, and 2. would enable us to build same as we do right now without the caching/recursive build feature.

andaaron: Also in the example above, in case 1, even without commit tags. There is no way to avoid building B, if it has not changed. If would be always built together with C.

tych0: > run the build for C, but because of the 'include' directive, the build would ignore the docker url to B, and use the one in the oci layout instead

I don't think we should do this though. My understanding of include: was just that it would allow you to specify type: built sources that weren't in the current file, but were in one of the files that was included. I don't think we should be re-writing layer sources on the fly, other than via --substitute.

andaaron: My understanding of the include is slightly different, it loosely specifies the order in which the files need to be built. I was wondering if include is really the right name. The name Include would also suggest the content of the stackerfiles is merged into a single list of layers before the build starts (in a single Stackerfile struct instance). But if we go this route the logic for caching stuff in a registry remotely would become even more complicated, and we’d also need to be extra careful with the relative imports.

Configuration names asside, if we’re not re-writing build sources, in order for layer caching to work, we have to upload B to the registry, to the same url specified in the stackerfile for C, before C is built. I guess we could do this by adding configuration for upload url and credentials for B. And use —substitute for the tags.

I would still use the —no-includes to just ignore the includes in the files. My concern is without it there wouldn’t be a way to build just one layer, we’d always have to built the whole tree and upload it to the registry. If we’re ok with rebuilding all layers specified in the includes, and rely on cache to skip the rebuild, then I can make the change.

tych0: > My understanding of the include is slightly different, it loosely specifies the order in which the files need to be built. I was wondering if include is really the right name. The name Include would also suggest the content of the stackerfiles is merged into a single list of layers before the build starts (in a single Stackerfile struct instance).

Yeah, it's possible that we need a different name. I was thinking that it would behave like an "include" when I suggested it, though, regardless of the actual implementation.

> Configuration names asside, if we’re not re-writing build sources, in order for layer caching to work, we have to upload B to the registry, to the same url specified in the stackerfile for C, before C is built.

I don't think we want to be doing intermediate uploads/downloads at all. What I'm suggesting is that stacker files with includes would look like:

```
/tmp/foo cat a.yaml 
a:
    from:
        type: docker
        url: docker://centos:latest
    run: ...
/tmp/foo cat b.yaml 
stacker_config:
    includes:
        - a.yaml
b:
    from:
        type: built
        tag: a
    run: ...
```

If we do it this way, I'm not sure an option called --no-includes really makes sense; the builds won't work without it. I'm afraid that doing some re-writing on the fly just seems too magical.

I understand that you want to do some build caching in a remote repo somewhere, but I think that's probably config that either lives in ~/.config/stacker.conf or is passed in via the command line, not necessarily a first class thing that users need to reason about when writing stacker files.

andaaron: I understand this model. But in order for incremental builds to work, we need to be able to skip the rebuild of 'a'  in your example, and pick it up from somewhere else. 

In your case let's assume both 'a' and 'b' import sources which need to be built or just copied in the image.

Now let's say for example yesterday I modify the sources imported in the stacker yaml of 'a', this means both 'a' and 'b' (which depends on 'a') needed to be rebuilt, and they were both rebuilt and uploaded to the remote registry to be consumed by various users.

Tomorrow, someone else, in a different build environment, modifies the sources imported in 'b'. 'a' does not have to be rebuilt since the imports have not changed since yesterday, and he should only have to wait for 'b' to be built. Yesterday's version of 'a' would be downloaded, and used as base for 'b'. 

So the build for 'b' needs to account both for the case when 'a' needs to be rebuilt, and the case in which 'a' doesn't need to be rebuilt. Always pushing 'a' to a repository, and optionally downloading it in case we need to skip it's build are part of the solution.

This is the kind of decision making we're currently doing in our Jenkins groovy + bash code. This is a real use case. 
And we need unify the build logic, so the build behaves the same way regardless if we build it in Jenkins or the developer does it on his individual machine.

If stacker is not the right place to make these kind of decisions, let's figure out which decisions should be made by stacker, and which should be done by other tools, be they in bash or go.

hallyn: > > My understanding of the include is slightly different, it loosely specifies the order in which the files need to be built. I was wondering if include is really the right name. The name Include would also suggest the content of the stackerfiles is merged into a single list of layers before the build starts (in a single Stackerfile struct instance).
> 
> Yeah, it's possible that we need a different name. I was thinking that it would behave like an "include" when I suggested it, though, regardless of the actual implementation.
> 
> > Configuration names asside, if we’re not re-writing build sources, in order for layer caching to work, we have to upload B to the registry, to the same url specified in the stackerfile for C, before C is built.
> 
> I don't think we want to be doing intermediate uploads/downloads at all. What I'm suggesting is that stacker files with includes would look like:
> 
> ```
> /tmp/foo cat a.yaml 
> a:
>     from:
>         type: docker
>         url: docker://centos:latest
>     run: ...
> /tmp/foo cat b.yaml 
> stacker_config:
>     includes:
>         - a.yaml
> b:
>     from:
>         type: built
>         tag: a
>     run: ...
> ```

The 'tag: a" should probably be "tag: a:a" ?  Else we're placing a new demand that all image builds in a repo have unique names.

tych0:
On Fri, Jun 07, 2019 at 09:55:39AM -0700, Serge Hallyn wrote:
> hallyn commented on this pull request.
> 
> 
> 
> > +	stackerFiles, err := NewStackerFiles(paths, opts.Substitute)
> +	if err != nil {
> +		return err
> +	}
> +
> +	// Initialize the DAG
> +	dag, err := NewStackerFilesDAG(stackerFiles)
> +	if err != nil {
> +		return err
> +	}
> +
> +	// In case we don't want to build the images found in includes, filter them out
> +	sortedPaths := dag.Sort()
> +	var finalPaths []string
> +	if !opts.NoIncludes {
> +		finalPaths = sortedPaths
> 
> > > My understanding of the include is slightly different, it loosely specifies the order in which the files need to be built. I was wondering if include is really the right name. The name Include would also suggest the content of the stackerfiles is merged into a single list of layers before the build starts (in a single Stackerfile struct instance).
> > 
> > Yeah, it's possible that we need a different name. I was thinking that it would behave like an "include" when I suggested it, though, regardless of the actual implementation.
> > 
> > > Configuration names asside, if we’re not re-writing build sources, in order for layer caching to work, we have to upload B to the registry, to the same url specified in the stackerfile for C, before C is built.
> > 
> > I don't think we want to be doing intermediate uploads/downloads at all. What I'm suggesting is that stacker files with includes would look like:
> > 
> > ```
> > /tmp/foo cat a.yaml 
> > a:
> >     from:
> >         type: docker
> >         url: docker://centos:latest
> >     run: ...
> > /tmp/foo cat b.yaml 
> > stacker_config:
> >     includes:
> >         - a.yaml
> > b:
> >     from:
> >         type: built
> >         tag: a
> >     run: ...
> > ```
> 
> The 'tag: a" should probably be "tag: a:a" ?  Else we're placing a new demand that all image builds in a repo have unique names.

No, I don't think so. We're placing the restriction that everything
included in a particular path needs to have a unique name, but I think
that's different.

build.go
tych0: > Now let's say for example yesterday I modify the sources imported in the stacker yaml of 'a', this means both 'a' and 'b' (which depends on 'a') needed to be rebuilt, and they were both rebuilt and uploaded to the remote registry to be consumed by various users.

> Tomorrow, someone else, in a different build environment, modifies the sources imported in 'b'. 'a' does not have to be rebuilt since the imports have not changed since yesterday, and he should only have to wait for 'b' to be built. Yesterday's version of 'a' would be downloaded, and used as base for 'b'.

I think we can do this in stacker, but I don't think this should be expected to be the default behavior. When you're first starting out with stacker, you're not going to have a setup as complex as we do, and you'll want to do includes, etc. just to play around with it, without necessarily uploading it. I think the default way to express things in the yaml should reflect this.

Now, we can definitely do something where we either accept via arguments, additional config in the build yaml, etc. a URL where potential builds might be cached (with a naming scheme dictated by stacker) so that stacker can try to resolve things if they've already been built. So I think we can still accomplish what you want, with additional configuration to make explicit what's happening, instead of implicitly rewriting stuff.

andaaron: In this case I'll just proceed with addressing the original comments in this PR. 
Will also remove the --no-includes directive.
And add tests.

Maybe change the name from 'includes' to 'prerequisites'.

andaaron:
> On Fri, Jun 07, 2019 at 09:55:39AM -0700, Serge Hallyn wrote: hallyn commented on this pull request. > +	stackerFiles, err := NewStackerFiles(paths, opts.Substitute) +	if err != nil { +	return err +	} + +	// Initialize the DAG +	dag, err := NewStackerFilesDAG(stackerFiles) +	if err != nil { +	return err +	} + +	// In case we don't want to build the images found in includes, filter them out +	sortedPaths := dag.Sort() +	var finalPaths []string +	if !opts.NoIncludes { +	finalPaths = sortedPaths > > My understanding of the include is slightly different, it loosely specifies the order in which the files need to be built. I was wondering if include is really the right name. The name Include would also suggest the content of the stackerfiles is merged into a single list of layers before the build starts (in a single Stackerfile struct instance). > > Yeah, it's possible that we need a different name. I was thinking that it would behave like an "include" when I suggested it, though, regardless of the actual implementation. > > > Configuration names asside, if we’re not re-writing build sources, in order for layer caching to work, we have to upload B to the registry, to the same url specified in the stackerfile for C, before C is built. > > I don't think we want to be doing intermediate uploads/downloads at all. What I'm suggesting is that stacker files with includes would look like: > > ``` > /tmp/foo cat a.yaml > a: > from: > type: docker > url: docker://centos:latest > run: ... > /tmp/foo cat b.yaml > stacker_config: > includes: > - a.yaml > b: > from: > type: built > tag: a > run: ... > ``` The 'tag: a" should probably be "tag: a:a" ? Else we're placing a new demand that all image builds in a repo have unique names.
> No, I don't think so. We're placing the restriction that everything included in a particular path needs to have a unique name, but I think that's different.

Tycho, Serge has a point here. We need to differentiate the images locally in the folders used by stacker. Should we include this in the code?

build.go
andaaron: Tycho,

Made some progress, but having some difficulty with how stacker handles the layers:
https://github.com/anuvu/stacker/pull/35/commits/f78ac09ac48e05920d59f1bf8cf1fea24ef3196b

Stacker assumes all layers are defined in the same stackerfile, and actively searches the stackerfile for the base layers. I've added some warnings instead of errors in a couple of places, but I don't know how to handles the one below.

It's the case when a layer depends on another layer of type  'BuiltType'.

Partial output from the new tests:
`# building image layer2...
# importing files...
# copying import2
# WARNING: base layer layer1_1 mising from stackerfile /home/andaaron/anuvu/stacker/ocibuilds/sub2/stacker.yaml!
# WARNING: base layer layer1_1 mising from stackerfile /home/andaaron/anuvu/stacker/ocibuilds/sub2/stacker.yaml!
# running commands...
# running commands for layer2
# + cp /stacker/import2 /root/import2
# + cp /root/import1 /root/import1_copied
# generating layer for layer2
# setting git version annotation to 9b158a7a4545c7dc022542cdb661acc746cbd984-dirty
# filesystem layer2 built successfully
# error: couldn't find a cache of base layer
`
Looks like it tries to look for the base layer in the cache, after the build, but it doesn't find it. The cache is coupled to the stackerfile, so not sure how to go about this.

build.go
andaaron: Added a command line flag '--order-only'

tych0:
> Tycho, Serge has a point here. We need to differentiate the images locally in the folders used by stacker. Should we include this in the code? If the user doesn't do it in the stacker.yaml(s) we need to take care of it internally.

Can you give an example of what you're talking about here? I'm not sure I understand. I think it's fine to just fail if there's some internal naming conflict, vs. fixing it up magically.

build.go
tych0: > The cache is coupled to the stackerfile, so not sure how to go about this.

I think that's the bug: in your test, you have sub1, sub2, sub3, but you only call the actual "stacker build" in sub3; stacker shouldn't write any files in any other directory besides sub3 in this case.

andaaron: I'm actually calling stacker outside those folders and sending the file as parameter with '-f'.

Is there an assumption the cache is always located relative to the current working directory?
In this case the bug should go away after I get rid of chdir.

andaaron:
> > Tycho, Serge has a point here. We need to differentiate the images locally in the folders used by stacker. Should we include this in the code? If the user doesn't do it in the stacker.yaml(s) we need to take care of it internally.
> 
> Can you give an example of what you're talking about here? I'm not sure I understand. I think it's fine to just fail if there's some internal naming conflict, vs. fixing it up magically.

Multiple stackerfiles may contain layers called 'build'. How do I know if the layer 'build' I'm using as base is the one in the current file, or in another file.

tych0:
> Multiple stackerfiles may contain layers called 'build'. How do I know if the layer 'build' I'm using as base is the one in the current file, or in another file.

Yep, I think it's safe to just fail in that case.

build.go
tych0: > I'm actually calling stacker outside those folders and sending the file as parameter with '-f'.
Seems like you probably want to pass --stacker-dir then, to make it whatever the original user passed as --stacker-dir (same for --oci-dir and --roots-dir, then). But it might be nice to try and do this all in golang, rather than fork+exec.

> Is there an assumption the cache is always located relative to the current working directory?
No, the cache is always located relative to --stacker-dir though.

andaaron: This is not the issue, the code already accounts for the --stacker-dir as an absolute path.

The issue seems to be at:
https://github.com/anuvu/stacker/blob/master/build.go#L679
which leads to
https://github.com/anuvu/stacker/blob/master/cache.go#L300
which leads to
https://github.com/anuvu/stacker/blob/master/cache.go#L281
which leads to
https://github.com/anuvu/stacker/blob/master/cache.go#L162

It's looking for a layer with the name of the base layer in the same stackerfile.

tych0: Ah, ok. Seems like that resolution will have to be updated then too.

andaaron: How to take care of it? If it was straight forward I wouldn't have asked :)
Also do you think it would make more sense to switch back to the name 'include', concatenate all layers in the same Stackerfile object, and call build() only once? Would that make more sense?

tych0: > How to take care of it?

I think the reason for needing the stackerfiles in the cache is so it can look up each individual layer, to figure out if the structure has changed (e.g. new run section, import, or bind mount). So I think you can just make OpenCache() take a []*Stackerfile instead, and it should just look through all of them sequentially.

> Also do you think it would make more sense to switch back to the name 'include', concatenate all layers in the same Stackerfile object, and call build() only once? Would that make more sense?

Oh, yeah, that may make things super easy. Just open them all, cat them all together, run substitutions, and parse the result? I guess you might have to do something slightly fancy with the stacker_config: section, but other than that it seems like maybe that would simplify all of this?

andaaron:
@tych0 @hallyn this PR is ready for another review.
Let me know if I need to add other tests.

base.go
```diff
@@ -316,18 +316,27 @@ func getOCI(o BaseLayerOpts) error {
...
 	// We need to copy any base OCI layers to the output dir, since they
 	// may not have been copied before and the final `umoci repack` expects
 	// them to be there.
 	base := o.Layer
 	for {
-		var ok bool
-		base, ok = sf.Get(base.From.Tag)
+		// Iterate through base layers until we find the first one with is not BuiltType
+		ok := false
+		for _, sf := range sfs {
```
tych0: This is essentially the same code as in the build cache's lookupLayerDefinition(); I wonder if you can make a typedef for a list of stackerfiles and just have the lookup as a method on that?

test/prerequisites.bats
```diff
@@ -0,0 +1,76 @@
...
+
+@test "order prerequisites" {
+    stacker build -f ocibuilds/sub3/stacker.yaml --order-only
+    [[ "${lines[-1]}" =~ ^(2 build .*ocibuilds\/sub3\/stacker\.yaml: requires: \[.*\/sub1\/stacker\.yaml .*\/sub2\/stacker\.yaml\])$ ]]
+    [[ "${lines[-2]}" =~ ^(1 build .*ocibuilds\/sub2\/stacker\.yaml: requires: \[.*\/sub1\/stacker\.yaml\])$ ]]
+    [[ "${lines[-3]}" =~ ^(0 build .*ocibuilds\/sub1\/stacker\.yaml: requires: \[\])$ ]]
+}
+
+@test "build layers and prerequisites" {
+    stacker build -f ocibuilds/sub3/stacker.yaml
```
tych0: Seems like it might be worth testing that e.g. the files exist in the right places after the builds?

andaaron: Done

test/basic.bats
```diff
@@ -69,7 +69,8 @@ function teardown() {
     [ "$(cat oci/blobs/sha256/$config | jq -r '.config.Entrypoint | join(" ")')" = "echo hello world" ]
 
     stackerGit=$(cat oci/blobs/sha256/$manifest | jq -r '.annotations."ws.tycho.stacker.git_version"')
-    [ "stacker version ${stackerGit}" = "$(stacker --version)" ]
+    rootDir=$(git rev-parse --show-toplevel)
+    [ "stacker version ${stackerGit}" = "$(${rootDir}/stacker --version)" ]
```
tych0: Can you describe what was broken here? AFAIK this should be a no-op.

andaaron: It was using the stacker in the PATH instead of the stacker in the local folder. So the versions were different.

tych0: It should really be using the function "stacker" from test/helpers.bash; can you describe how to reproduce it? This has worked for me for quite a while...

andaaron: Does `$()` for with functions?

tych0:
```
~ function foo() { echo "hello world"; } 
~ echo "$(foo)"
hello world
```

Yep. Can you describe how to reproduce the problem you were seeing?

andaaron:
```
andaaron@andaaron-ThinkPad-P51s:~/anuvu/stacker$ sudo bats test/basic.bats --tap
[sudo] password for andaaron: 
1..1
not ok 1 basic workings
# (in test file test/basic.bats, line 73)
#   `[ "stacker version ${stackerGit}" = "$(stacker --version)" ]' failed
# Getting image source signatures
# Copying blob sha256:8ba884070f611d31cb2c42eddb691319dc9facf5e0ec67672fcfa135181ab3df
# Copying config sha256:b9a1b1f0b2baaec83946a26d7045e4028f11eccc8b0e5b3514568e56a391beb2
# Writing manifest to image destination
# Storing signatures
# initializing stacker recipe: stacker.yaml
# substituting $FAVICON to favicon.ico
# stacker build order:
# 0 build /home/andaaron/anuvu/stacker/test/stacker.yaml: requires: []
# building: 0 /home/andaaron/anuvu/stacker/test/stacker.yaml
# substituting $FAVICON to favicon.ico
# building image centos...
# importing files...
# copying /home/andaaron/anuvu/stacker/test/stacker.yaml
# downloading https://www.cisco.com/favicon.ico
 1.37 KiB / 1.37 KiB [========================================================] 100.00% 2.03 MiB/s 0s
# copying /home/andaaron/anuvu/stacker/test/executable
# using cached copy of .stacker/layer-bases/centos.tar
# running commands...
# running commands for centos
# + cp /stacker/favicon.ico /favicon.ico
# + cp /stacker/executable /usr/bin/executable
# generating layer for centos
# setting git version annotation to 7aff18162f3533559a6b466f10d6f2faa71da6eb-dirty
# filesystem centos built successfully
# building image layer1...
# importing files...
# running commands...
# running commands for layer1
# + rm /favicon.ico
# generating layer for layer1
# setting git version annotation to 7aff18162f3533559a6b466f10d6f2faa71da6eb-dirty
# filesystem layer1 built successfully
andaaron@andaaron-ThinkPad-P51s:~/anuvu/stacker$ stacker --version
stacker version e5283a6aceab5f3ca795bf80c2399d8c91f5a8b0-dirty
andaaron@andaaron-ThinkPad-P51s:~/anuvu/stacker$ ./stacker --version
stacker version 7aff18162f3533559a6b466f10d6f2faa71da6eb
andaaron@andaaron-ThinkPad-P51s:~/anuvu/stacker$ git rev-parse HEAD
7aff18162f3533559a6b466f10d6f2faa71da6eb
```

andaaron: Nevermind, it was because I didn't rebuild stacker. And the new binary should have included 'dirty' in the name.

andaaron: @tych0 so should the version in the manifest be the version of the stacker binary used for building?
Or should it be the version of the repo where the stacker yaml is located?

andaaron: Removed this commit

base.go
andaaron: Done, but with a map, since we were already using a map in other places, and right now we have consistent usage.

test/basic.bats
tych0: It's the repo where the stacker.yaml is located; they just happen to be the same for the actual stacker repo.

andaaron:
Done with fixing last typos and squashes.

tych0:
Great, thanks!

andaaron:
@tycho, for the next step, building multiple stackerfiles under a particular repo.

I see two possible options at the moment.
a) Stacker receives a file containing a list of stackerfiles as input, stacker would still compute the right build order, and build the dependencies even if not already specified in the file
b) Search and build all stackerfiles under the current working directory which have paths matching a regex

Option a) would give more control to the user, but would require extra maintenance for the new file.
Option b) would require less maintenance, but would require extra care with naming the files, and in what paths they are placed

What do you think?

tycho:
You've tagged the wrong guy. You wanted @tych0.

andaaron:
> You've tagged the wrong guy. You wanted @tych0.

Yeah, sorry about that

PR #38, opened 2019-06-28
Allow building of multiple stackerfiles sent from command line with '-f' 

andaaron:
This is for building stackerfiles which do not depend on each other, but may have the same base images.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -16,10 +16,9 @@ var buildCmd = cli.Command{
 			Name:  "leave-unladen",
 			Usage: "leave the built rootfs mount after image building",
 		},
-		cli.StringFlag{
+		cli.StringSliceFlag{
 			Name:  "stacker-file, f",
 			Usage: "the input stackerfile",
-			Value: "stacker.yaml",
 		},
 		cli.BoolFlag{
 			Name:  "no-cache",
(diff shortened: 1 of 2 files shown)


tych0:
Is there a reason to do this vs. just doing multiple stacker invocations? Shouldn't the caching figure out that the base images are the same, so the work will only be done once?

andaaron:
Having a loop in a separate script needs extra maintenance.  
Currently we're maintaining such scripts on 3 separate repos, and they all have similar logic. I was thinking of removing them entirely.

tych0:
Don't you still need code to figure out where the stacker files are, though? Would it be less work to have a "stacker build ./..." type of thing?

andaaron:
The other option was something similar 'stacker build ./...' but this seems more explicit as:
- 'stacker build ./...' would only look for file names 'stacker.yaml', and we'd also have to add an extra parameter for a pattern of the file name
- there will also need to be a file naming convention inside the folder for files to match or not match this pattern. The developers may want some files to be built (for example images they want to publish), but not others (for example run a binary build inside a container and only keep the resulting binary).
- I haven't found a reliable library for matching glob patterns with '**' matching paths containing '/'. We could use a regex for finding files in the current working directory (or we could have an extra parameter for the file names, beside the pattern)
- If we're not going for a glob/regex, how would a './...' style pattern match the needed file names?

tych0:
> 'stacker build ./...' would only looks for file names 'stacker.yaml', and we'd also have to add an extra parameter for a pattern of the file name

Yes, but in practice most (all?) of our stacker files for situations where we'd use this are called stacker.yaml.

> there would also need to be a file naming convention inside the folder for files to match or not match this pattern.

Yes, this is a fair point. The way we currently have things organized, it would build a bunch of stuff that we don't necessarily need.

Issue #2, opened 2018-01-30
Don't hard-code uid 100000 for the container uid mapped to your host uid


hallyn:
It would be best to calculate the maxuid mapped in the container and add one to that.

Otherwise unfortunate users who have a subuid allocation of > 100000 uids will get failures with unprivileged stacker runs.

tych0:
My bad, thanks!

PR #39, opened 2019-06-30
Add the option to save the new layers to a remote location

andaaron:
Feature can be useful if the remote is a docker registry, but can also work with an OCI layout as "remote".
Using an OCI layout is currently useful for testing purposes.

### DIFF
--- a/build.go
+++ b/build.go
@@ -9,12 +9,10 @@ import (
 	"os/exec"
 	"os/user"
 	"path"
-	"path/filepath"
 	"runtime"
 	"strings"
 	"time"
 
-	stackeroci "github.com/anuvu/stacker/oci"
 	"github.com/openSUSE/umoci"
 	"github.com/openSUSE/umoci/mutate"
 	"github.com/openSUSE/umoci/oci/casext"
@@ -23,6 +21,9 @@ import (
 	"github.com/pkg/errors"
 	"github.com/vbatts/go-mtree"
 	"golang.org/x/sys/unix"
+
+	"github.com/anuvu/stacker/lib"
+	stackeroci "github.com/anuvu/stacker/oci"
 )
 
 type BuildArgs struct {
@@ -35,6 +36,7 @@ type BuildArgs struct {
 	LayerType               string
 	Debug                   bool
 	OrderOnly               bool
+	RemoteSaveTags          []string
 }
 
 func updateBundleMtree(rootPath string, newPath ispec.Descriptor) error {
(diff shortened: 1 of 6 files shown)


andaaron:
Note this PR is just for saving the layers to be used later.

I intend to add another command line flag "--remote-caching" which would enable retrieving the layers from dockerhub (or the oci layout in case of tests) if they have no changes in Git since the commit in 'commit-<id>'.

With both these changes we should be able to obsolete the logic we have in our groovy code which cannot be executed outside Jenkins.

build.go
```diff
@@ -313,6 +316,63 @@ func generateSquashfsLayer(oci casext.Engine, name string, author string, opts *
...
+	if len(tags) == 0 {
+		fmt.Printf("can't save layer %s since list of tags is empty\n", name)
+	}
+
+	// Store the layers to new detination
+	for _, tag := range tags {
+		var destUrl string
+		switch is.Type {
+		case DockerType:
+			destUrl = fmt.Sprintf("%s/%s:%s", strings.TrimRight(sf.buildConfig.SaveUrl, "/"), name, tag)
```
tych0: I think `path.Join()` would get rid of the need for the `strings.TrimRight()` bit? Not sure if it looks any better or not.

build.go
```diff
@@ -313,6 +316,63 @@ func generateSquashfsLayer(oci casext.Engine, name string, author string, opts *
...
 }
 
+// SaveLayer stores the final layers into a separate location based on the content of
+// the stackerfile, this is useful to avoid an extra manual step to upload build results
+// and also in case of caching in between stacker builds
+// The logic should work for both Docker registry destination and OCI layout destinations
+// In case of OCI layout destinations the tag will be included in the layer name
+func SaveLayer(opts *BuildArgs, sf *Stackerfile, name string) error {
+	if len(sf.buildConfig.SaveUrl) == 0 {
+		fmt.Printf("layer %s will not be save since it doesn't have a save URL\n", name)
```
tych0: Can you drop this? It sounds scary, and it seems like mostly normal operation.

cmd/build.go
```diff
@@ -66,6 +74,13 @@ func beforeBuild(ctx *cli.Context) error {
 		}
 	}
 
+	if !ctx.IsSet("remote-save") {
+		if len(ctx.StringSlice("remote-save-tag")) != 0 {
+			// Cannot set upload tags if uploading to storage URL is disabled
+			return fmt.Errorf("--remote-save-tag requires --remote-save")
+		}
+	}
+
```
tych0: Seems like we also need a check for when you supply `--remote-save` but forget `--remote-save-tag`; or better yet, can we just have `--remote-save-tag` and drop `--remote-save` all together? You'd have to supply the full URL, but then you could save to multiple repositories, which might be nice.

git.go
```diff
@@ -5,26 +5,84 @@ import (
...
+
+	// Determine git username
+	username, err := GitUsername(path)
+	if err != nil {
+		return "", err
+	}
+
+	// Set username in tag
+	if len(username) != 0 {
+		tag = username + "-"
```
tych0: Won't this defeat the ability for you to use caches that others have built?

build.go
andaaron: Does that work for URLs? I though it's just for file paths on the disk.

build.go
andaaron: This function is not be called in normal operations, only if you use the command line option --remote-save.
Should I remove it?

git.go
andaaron: It would, but we can't have everyone push 'commit-id-dirty', they would be stepping on each other's toes.

cmd/build.go
andaaron: There's 4 cases right now:
1. File is in Git and --remote-save-tag is sent -> both commit-id tag and the tag from command line are used 
2. File is in Git and  --remote-save-tag is not sent -> only the commit-id tag is used
3. File is not in Git and --remote-save-tag is sent -> only the tags from command line are used
4. File is not in Git and the --remote-save-tag is not sent is not set -> the image is not saved

Because of #2, if we drop the --remote-save it means we always try to save with commit-id, and that would mean there's no control on how to disable the feature, unless the user edits all the relevant stacker.yaml files locally.

Please explain the part about having to supply the full URL. Are you suggesting we should send it from command line, and only if we send it from command line we save? 
In this case how do we distinguish between images we want to save and the others?

git.go
tych0: I would say don't push -dirty commits at all; who knows what's in them.

build.go
tych0: You're right that it is sort of an abuse of `path.Join()`; it works just fine though, `path.Join()` doesn't actually look at the filesystem at all when doing work. But feel free to ignore this if you feel like it's a hack compared to what you've got now.

build.go
tych0: Oh, right. I think this is really just an implementation of the check I wanted below; can we do it earlier though so we can fail faster (or just refactor it out per discussion below)?

cmd/build.go
tych0: Sorry, you're right. I think we can just get rid of `--remote-save` entirely, then, and if you've supplied a URL then you get the caching, and if you've supplied `--remote-save-tag`, you get additional tags. Does that sound reasonable?

build.go
andaaron: Not sure path.Join is ok for this, afaik it calls path.Clean which removes double slashes, and may remove the double slashes separating the protocol from the hostname.

git.go
andaaron: Done

cmd/build.go
andaaron: Done

build.go
andaaron: Done

PR #48, opened 2019-09-10
Add an option to search for stackerfiles under a given folder

andaaron:


cmd/build.go
```diff
@@ -80,6 +89,21 @@ func beforeBuild(ctx *cli.Context) error {
 		return fmt.Errorf("unknown layer type: %s", ctx.String("layer-type"))
 	}
 
+	if len(ctx.String("search-dir")) > 0 {
```
tych0: I wonder if this shouldn't be a separate command, `stacker recursive-build ./` or similar. What do you think? The code itself looks good.

andaaron: We could do that. Thing is the normal build was recursive anyway since #35 .
This can be seen in the existing "order prerequisites" test case.

I've talked with @rchamarthy today, and he want to review all this work, including #35 and #39 .
Let's setup a meeting.

tych0:
Awesome, thanks!

PR #49, opened 2019-09-11
fix the annotations to use com.github.anuvu

rchamarthy:
Signed-off-by: Ravi Chamarthy <ravchama@cisco.com>

### DIFF
--- a/api.go
+++ b/api.go
@@ -19,8 +19,8 @@ import (
 
 const (
 	MediaTypeImageBtrfsLayer  = "application/vnd.cisco.image.layer.btrfs"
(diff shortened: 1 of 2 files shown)


tych0:
Shouldn't it really be com.cisco?

tych0:
> making it cisco will make it proprietary. Its best the tags be from the tool as well.

Huh? The license makes it proprietary (or not), not the annotations. This feels incorrect, since we're invading github's domain space (maybe if we did io.github instead, but even then...). I think we should use com.cisco or buy a domain for the tool and use that instead.

rchamarthy:
@kprav33n dont we have a domain for aci or atomix that you bought? Can i use that here?

PR #52, opened 2019-09-13
Feature/shells other than sh

smoser:


smoser:
Just looking at one of our stacker files, and saw another reason why this is useful. It had a `set -o pipefail`, which is actually a bash-ism.  So it is relying on the fact that /bin/sh is a symlink to /bin/bash in the container that it is building in.  And quite possibly it is even a bug in bash as when it is invoked as 'sh' it generally tries to behave in more "posix" way.

build.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/openSUSE/umoci/pkg/fseval"
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
-	"github.com/vbatts/go-mtree"
+	mtree "github.com/vbatts/go-mtree"
```
tych0: The package is actually called 'mtree'; what is your editor doing here? If we're going to do this, we should do it tree-wide, but it seems like it shouldn't be necessary.

tych0:
The other patch looks fine to me, so let's figure out the mtree thing and I'm happy to merge this.

smoser:
Well, poked around a bit. [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) is what is re-writing the import line.  Heres a simple example that uses mtree, and then the 'goimports' rewrites it.

    $ cat f.go 
    package stacker

    import (
    	"os"
    	"github.com/vbatts/go-mtree"
    )
    
    func testMe(fh os.File) error {
    	_, err := mtree.ParseSpec(fh)
    	return err
    }

Then running goimports on it


    $ goimports f.go 
    package stacker
    
    import (
    	"os"
    
    	mtree "github.com/vbatts/go-mtree"
    )
    
    func testMe(fh os.File) error {
    	_, err := mtree.ParseSpec(fh)
    	return err
    }

tych0:
Might be worth filing a bug with goimports then?

smoser:
well... so i will drop the goimports change.

Just for my own future reference.

    $ cat f.go
    package main
    
    import (
    	"io"
    
    	"github.com/vbatts/go-mtree"
    )
    
    func testMe(r io.Reader) error {
    	_, err := mtree.ParseSpec(r)
    	return err
    }
    
    func main() {
    	return
    }

    $ goimports f.go | grep mtree
    	_, err := mtree.ParseSpec(r)

    $ go get github.com/vbatts/go-mtree

    $ goimports f.go | grep mtree
    	"github.com/vbatts/go-mtree"
    	_, err := mtree.ParseSpec(r)


So, in conclusion, its the lack of goimports being able to be find in the search path that caused that.

smoser:
dropped the unrelated change to the imports.merge away.

tych0:
Right, thanks!

PR #53, opened 2019-09-22
Add stacker publish

andaaron:
`stacker publish` copies layers defined in stacker yaml files to a docker registry or oci layout

### DIFF
--- a/publisher.go
+++ b/publisher.go
@@ -3,8 +3,12 @@ package stacker
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
+
+	"github.com/openSUSE/umoci"
+	"github.com/openSUSE/umoci/oci/casext"
 
 	"github.com/anuvu/stacker/lib"
 )
@@ -18,6 +22,7 @@ type PublishArgs struct {
 	Url        string
 	Username   string
 	Password   string
+	Force      bool
 }
 
 // Publisher is responsible for publishing the layers based on stackerfiles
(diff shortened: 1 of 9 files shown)


publisher.go
```diff
@@ -0,0 +1,150 @@
...
+		for _, tag := range tags {
+			// Determine full destination URL
+			var destUrl string
+			switch is.Type {
+			case DockerType:
+				destUrl = fmt.Sprintf("%s/%s:%s", strings.TrimRight(opts.Url, "/"), name, tag)
+			case OCIType:
+				destUrl = fmt.Sprintf("%s:%s_%s", opts.Url, name, tag)
+			default:
+				return fmt.Errorf("can't save layers to destination type: %s", is.Type)
```
tych0: We will probably need zot:// here shortly. Happy to merge it without zot for now, but can you do a patch?

publisher.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	// Need to determine if URL is docker/oci or something else
+	is, err := NewImageSource(opts.Url)
+	if err != nil {
+		return err
+	}
+
+	// Iterate through all layers defined in this stackerfile
+	for _, name := range sf.fileOrder {
```
tych0: What do you think about checking the cache before we push? Just in case something has changed? Or should we not do that?

publisher.go
andaaron: Are zot:// urls similar to docker urls? Same except for protocol?

publisher.go
andaaron: Added a check. By default we verify cache, but I also added a '--force' command line option for override this.

publisher.go
andaaron: Made the changes. Also found another place where zot:// is missing.

andaaron:
@tych0 Made the changes. 

Though I'm not entirely happy with the UX and the flags/validation. Suggestions are welcomed.

PR #54, opened 2019-10-09
Feature/passthrough environment

smoser:
naming is hard. input on `container_env_passthrough` and `container_env` names is accepted.


tych0:
Cool, thanks!

PR #56, opened 2019-10-24
Stacker should do a HTTP HEAD call check the Content-Length and X-Checksum-Sha256 of the files imported using HTTP.

andaaron:
Fix for https://github.com/anuvu/stacker/issues/44
    
    There are several cases:
    - If the file is not in cache download it - covered by tests
    - If the file is in cache but there's an error connecting to the server skip the download - covered by tests
    - If the file is in cache and the shasum is present in HTTP head and matching the file cached locally skip the download - NOT covered by tests
    - If the file is in cache and the content length in HTTP head matches the size of the file cached locally skip the download - covered by tests
    - If the file is in cache and both the shasum and the content length are different from the file cached locally download the file - NOT covered by tests

### DIFF
--- a/network.go
+++ b/network.go
@@ -4,8 +4,11 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"path"
+	"strconv"
+	"strings"
 
 	"github.com/cheggaaa/pb"
 )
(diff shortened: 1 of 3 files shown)


andaaron:
Note not all servers expose the hash of the file in HTTP head calls. No way to test that part in the bats scripts without an Artifactory sever or a similar server.

test/import-http.bats
```diff
@@ -0,0 +1,70 @@
...
+    stacker build -f img/stacker.yaml
+    umoci unpack --image oci:img dest/img
+    [ "$(sha reference/favicon_orig.ico)" == "$(sha .stacker/imports/img/favicon.ico)" ]
+    [ "$(sha reference/favicon_orig.ico)" == "$(sha dest/img/rootfs/root/favicon.ico)" ]
+}
+
+@test "importing cached file from http url" {
+    mkdir -p .stacker/imports/img
+    cp reference/favicon_orig.ico .stacker/imports/img/favicon.ico
+    stacker build -f img/stacker.yaml
```
tych0: Seems like you should grep stacker's output here to make sure it actually uses the cache? Otherwise this doesn't really test anything, right?

network.go
```diff
@@ -4,29 +4,67 @@ import (
...
-		} else {
+		} else if localSize == remoteSize {
+			// Cached file has same content length as the remote file
+			fmt.Println("matched content length of", url)
+			fmt.Println("taking a leap of faith using cached copy of", url)
+			return name, nil
+		}
+		// Cached file has a different hash from the remote one
+		// Need to cleanup
+		err = os.RemoveAll(name)
```
andaaron: Is this necessary?

PR #58, opened 2019-11-14
Use git describe to calculate a version.

smoser:
Rather than spitting out a version that is simply a commit hash,
use output from git-describe.

Before:
  $ stacker --version
  stacker version 982c0032ce1a2cebf45d31904d920f20cabaa549

After
  $ stacker --version
  stacker version v0.4.0-78-gb16a7a1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,10 @@
 GO_SRC=$(shell find . -name \*.go)
-COMMIT_HASH=$(shell git rev-parse HEAD)
-COMMIT=$(if $(shell git status --porcelain --untracked-files=no),$(COMMIT_HASH)-dirty,$(COMMIT_HASH))
+VERSION=$(shell git describe --tags || git rev-parse HEAD)
+VERSION_FULL=$(if $(shell git status --porcelain --untracked-files=no),$(VERSION)-dirty,$(VERSION))
 TEST?=$(patsubst test/%.bats,%,$(wildcard test/*.bats))
 
 stacker: $(GO_SRC)
-	go build -ldflags "-X main.version=$(COMMIT)" -o stacker ./cmd
+	go build -ldflags "-X main.version=$(VERSION_FULL)" -o stacker ./cmd
 
 # make test TEST=basic will run only the basic test.
 .PHONY: check
--- a/git.go
+++ b/git.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"os/exec"
 	"strings"
+
+	"github.com/apex/log"
 )
 
 // gitHash generates a version string similar to git describe --always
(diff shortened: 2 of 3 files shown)


tych0:
Looks like the test needs a fix, but other than that seems fine to me.

smoser:
I'm pretty sure the error comes from test/basic.bats:

    stackerGit=$(cat oci/blobs/sha256/$manifest | jq -r '.annotations."ws.tycho.stacker.git_version"')
    [ "stacker version ${stackerGit}" = "$(stacker --version)" ]

Thoughts on how to fix that?

tych0:
Oh, right, this is where the comment was for :). Looks like git.go:GitVersion() needs updated as well.

smoser:
just for some context, part of an irc discussion.

<smoser> tycho: for the annotation on a oci, i think the git hash is fine.
<smoser> even preferable
<smoser> well, at least fine.
<smoser> i could argue that either way
<smoser> but i'm arguing 'stacker --version' having the v0.4.0-NN-gabcde is beneficial
<tycho> smoser: i don't understand why you'd want one the one way, and one the other?
<smoser> maybe i'm mistaken on the use of the annotations
<smoser> but i assumed they were just additional information about a oci blob
<tycho> yep, they are
<smoser> not specifically about stacker
<tycho> basically just, if you were in a git repo at the time of building this OCI blob, here's the git version
<tycho> yeah, that's right
<smoser> in which case, the git repo that you happen to be building a oci blob from may not have any tags
<tycho> right
<tycho> but it also might have tags
<smoser> i dont know. it feels to me like the full git hash is more useful there.
<smoser> git describe often needs "help" to do the right thing
<smoser> and stacker isn't necexsarily able to know what is right for eaach git repo it happens to come across
<smoser> specifically, often times someone would want a '--exclude' or '--match', or not *want* --tags (requiring a signed tag to be considered)
<smoser> git describe is not ever *wrong* in those cases (as you can still rely on the gHASH portion)
<smoser> but it may produce odd or confusing strings for the version portion of its output

smoser:
What is here right now is to continue using the full git hash to describe image repos, but use 'git describe' to describe stacker.

if Tycho wants to insist on using `git describe` everywhere, then i can adjust.

smoser:
11:57 <smoser> if you wanted to land https://github.com/anuvu/stacker/pull/58
11:57 <smoser> then i'd get that.
11:58 <smoser> but if you dont want to land it just as it is, then i'm ok with that, and I am willing to make the containers get marked with 'git describe' also.
12:03 <@tycho> smoser: yeah, i guess i'd rather keep it the same in both cases



so... /me goes to appease the maintainer.

smoser:
@tych0 ,

I've adjusted to your request.

PR #59, opened 2019-11-25
Add validation for stackerfile values

andaaron:
In this case base images of type 'built' need to have a 'tag' which is not empty


tych0:
Cool, thanks!

PR #60, opened 2019-12-07
from:scratch

hallyn:
Provide an example using from:scratch.

### DIFF
--- a/build.go
+++ b/build.go
@@ -349,16 +349,17 @@ func (b *Builder) Build(file string) error {
 		}
 
 		if len(run) != 0 {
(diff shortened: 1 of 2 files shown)


rchincha:
serge, you just had to do this ... :)

hallyn:
> serge, you just had to do this ... :)

Well there's a valid question of whether we want to drop this altogether in favor of just filling in a tarball in the build container and making the final container from: type: tar url: //stacker/build/tarball.tar.

Though if we drop from:scratch and drop the second patch in this pr, then the first patch technically is unrelated and still needed.

tych0:
This LGTM (I believe the test failure is because of a regression in golang 13.5, investigating more now).

r.e. scratch, it seems sort of painful to use this way, so maybe we should just drop it? Not sure.

PR #61, opened 2019-12-10
base: fix incorrect importing of base layers

tych0:
The goal of the import here is to get whatever OCI stuff that already
exsits into the output image, so that we can add the new stuff on top of
this.

We were doing this incorrectly: we don't want to go to the first non-built
layer, in fact we want the closest reasonable layer to what's there now. So
instead of doing a recursive look through the layer definitions, let's look
at the base layer's output, if it doesn't exist, look for its input, and if
that doesn't exist, give up and put the whole rootfs in the next OCI
output.

Signed-off-by: Tycho Andersen <tycho@tycho.ws>

### DIFF
--- a/base.go
+++ b/base.go
@@ -348,50 +348,47 @@ func getBuilt(o BaseLayerOpts, sfm StackerFiles) error {
 	// may not have been copied before and the final `umoci repack` expects
 	// them to be there.
 	targetName := o.Name
-	base := o.Layer
-	for {
-		// Iterate through base layers until we find the first one which is not BuiltType
-		// Need to declare ok separately, if we do it in the same line as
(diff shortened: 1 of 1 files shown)


tych0:
It keeps track of it by the tag it uses in the output image. It uses the name of the image to be built, and then updates that tag later with the mutated contents from the run: section.

tych0:
If that seems sort of hacky it's because it probably is. I wouldn't mind a patch that moved all of this "output" until right at the end when we actually need it :)

andaaron:
> It keeps track of it by the tag it uses in the output image. It uses the name of the image to be built, and then updates that tag later with the mutated contents from the run: section.

OK, got it.

PR #63, opened 2019-12-11
In case of build-only layers, look for base layers in OCIDir before looking into layer-bases

andaaron:
Fix bug introduced in: https://github.com/anuvu/stacker/commit/dce7c10c86b4c67f5cd54ce2c7e8da43227420c1
If a build-only layer depended on a non-build-only layer which was previously built (not downloaded), it would try to copy the base image from 'layer-bases' instead of the OCIDir where stacker places it

### DIFF
--- a/base.go
+++ b/base.go
@@ -366,6 +366,17 @@ func getBuilt(o BaseLayerOpts, sfm StackerFiles) error {
 		// it was sourced from e.g. a tar file, in which case there's
 		// nothing we can do besides initialize an empty oci tag and
 		// generate the whole thing.
+
+		// Attempt to copy the base from o.Config.OCIDir layout in case it's already there
+		err = lib.ImageCopy(lib.ImageCopyOpts{
+			Src:  fmt.Sprintf("oci:%s:%s", o.Config.OCIDir, baseInputTag),
+			Dest: fmt.Sprintf("oci:%s:%s", o.Config.OCIDir, targetName),
+		})
+		if err == nil {
+			return nil
+		}
+
+		// Attempt to copy the base from the layer-bases layout
 		cacheDir := path.Join(o.Config.StackerDir, "layer-bases", "oci")
 		err = lib.ImageCopy(lib.ImageCopyOpts{
 			Src:  fmt.Sprintf("oci:%s:%s", cacheDir, baseInputTag),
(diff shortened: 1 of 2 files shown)


base.go
```diff
@@ -366,16 +366,30 @@ func getBuilt(o BaseLayerOpts, sfm StackerFiles) error {
...
+
+		// Attempt to copy the base from o.Config.OCIDir layout in case it's already there
+		err = lib.ImageCopy(lib.ImageCopyOpts{
+			Src:  fmt.Sprintf("oci:%s:%s", o.Config.OCIDir, baseInputTag),
+			Dest: fmt.Sprintf("oci:%s:%s", o.Config.OCIDir, targetName),
+		})
+		if err == nil {
+			return nil
+		}
+		fmt.Printf("warning: %s", err)
```
tych0: I'd really rather not print these errors if we expect them. I probably set a bad example by my "just give up" comment.

But in this particular case, I'm not sure what this code does. Isn't it always not there? I don't think this should ever succeed.

base.go
```diff
@@ -366,16 +366,30 @@ func getBuilt(o BaseLayerOpts, sfm StackerFiles) error {
...
-			// give up and just create an empty image
-			return umociInit(o)
+		if err == nil {
+			return nil
 		}
-		return nil
+		fmt.Printf("warning: %s", err)
+
+		// Give up and just create an empty image
+		return umociInit(o)
```
tych0: If we drop the printf, I think we can drop this whole hunk, since it seems like a no-op?

andaaron: Is there a case for the image to be missing from both 'layer-bases' and 'o.Config.OCIDir'? If there is we still need this?

tych0: Yes, I think so. For example, if you import a tar, for a build_only layer, but then base a layer that we do actually export on it later.

andaaron: In this case I don't get the question, what part can be dropped?

tych0: This hunk just rearranges:
```
if err != nil {
    return umociInit(o)
}
```
to
```
if err == nil {
    return nil
}
return umociInit(o)
```
which I believe is a no-op.

andaaron: Reverted. I had rewritten it for consistency with the other copy. Where I wanted to avoid a nested if.

tych0:
I believe this is fixed by the very recently merged #64

PR #64, opened 2019-12-12
Fix for building layers based on build-only layers based on non-build-only layers

andaaron:


base.go
```diff
@@ -348,47 +348,51 @@ func getBuilt(o BaseLayerOpts, sfm StackerFiles) error {
...
-		if err != nil {
-			// give up and just create an empty image
-			return umociInit(o)
+		if base.From.Type != BuiltType || !base.BuildOnly {
+			break
 		}
+	}
+
+	// Nothing to do here -- we didn't import any base layers.
+	if base.From.Type != DockerType && base.From.Type != OCIType && base.From.Type != ZotType && base.From.Type != BuiltType {
```
tych0: I think this is still wrong, we always have to do *something* here for the squashfs case, because the resulting code updates a tag it expects to exist.

base.go
```diff
@@ -348,47 +348,56 @@ func getBuilt(o BaseLayerOpts, sfm StackerFiles) error {
...
+			break
 		}
+	}
 
+	if baseType != DockerType && baseType != OCIType && baseType != ZotType {
+		// Assume the user who wrote the stacker yaml has ordered the layers correctly
+		// and the base image has already been built and placed in  OCIDir
 		refs, _ := o.OCI.ListReferences(context.Background())
 		fmt.Printf("refs: %v\n", refs)
 		return lib.ImageCopy(lib.ImageCopyOpts{
```
tych0: If the layer is build-only, this fails, right?

base.go
```diff
@@ -348,47 +348,56 @@ func getBuilt(o BaseLayerOpts, sfm StackerFiles) error {
...
+		if !base.BuildOnly {
+			break
 		}
+	}
 
+	if baseType != DockerType && baseType != OCIType && baseType != ZotType {
+		// Assume the user who wrote the stacker yaml has ordered the layers correctly
+		// and the base image has already been built and placed in  OCIDir
 		refs, _ := o.OCI.ListReferences(context.Background())
 		fmt.Printf("refs: %v\n", refs)
```
tych0: We can drop this debug

base.go
andaaron: If the only layers which can be 'build-only' are BuiltType, then this should be covered.
The look should break only if base is not BuiltType (line 370) or it is BuiltType but it is not 'build-only' (line 379)

tych0:
Great, thanks!

PR #68, opened 2020-01-29
Remove automatic tagging of uploaded layers with commit-<id>

andaaron:


tych0:
Can you add a bit to the commit message about why this is unneeded?

andaaron:
> Can you add a bit to the commit message about why this is unneeded?

Done.

PR #71, opened 2020-02-01
import: remove the proper path

### LINKED ISSUE #70: caching issue in stacker build
1. import a dir with one file in it in a stacker.yaml file
2. do a stacker build
3. touch a file in that dir
4. do a stacker build
5. remove the file in that you additionally created
6. do a stacker build, it reports found a cached layer
7. In fact it found the wrong cached layer, when you unpack you see the file you created in step 3 in the layer

hallyn:
Closes #70

When a file under an imported directory is removed, calculate the
correct path to remove.

### DIFF
--- a/import.go
+++ b/import.go
@@ -120,9 +120,9 @@ func importFile(imp string, cacheDir string) (string, error) {
 	for _, d := range diff {
 		switch d.Type() {
 		case mtree.Missing:
-			err := os.RemoveAll(path.Join(cacheDir, d.Path()))
+			err := os.RemoveAll(path.Join(cacheDir, path.Base(imp), d.Path()))
 			if err != nil {
-				return "", errors.Wrapf(err, "couldn't remove missing import %s", path.Join(cacheDir, d.Path()))
+				return "", errors.Wrapf(err, "couldn't remove missing import %s", path.Join(cacheDir, path.Base(imp), d.Path()))
 			}
 		case mtree.Modified:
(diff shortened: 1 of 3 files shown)


import.go
```diff
@@ -120,7 +120,7 @@ func importFile(imp string, cacheDir string) (string, error) {
 	for _, d := range diff {
 		switch d.Type() {
 		case mtree.Missing:
-			err := os.RemoveAll(path.Join(cacheDir, d.Path()))
+			err := os.RemoveAll(path.Join(cacheDir, path.Base(imp), d.Path()))
 			if err != nil {
 				return "", errors.Wrapf(err, "couldn't remove missing import %s", path.Join(cacheDir, d.Path()))
```
tych0: Can you fix the path in the error message too? Otherwise lgtm.

andaaron:
Maybe we should add a test which reproduces the conditions in which the bug was found.

hallyn:
> Maybe we should add a test which reproduces the conditions in which the bug was found.

Oh, actually there was a test, but it was wrong.  Trying to fix it.  But man bats, is annoying to debug.

hallyn:
> > Maybe we should add a test which reproduces the conditions in which the bug was found.
> 
> Oh, actually there was a test, but it was wrong. Trying to fix it. But man bats, is annoying to debug.

sorry, yes, i did find it later that night.

PR #74, opened 2020-02-14
unpriv-setup: handle when our uid is in the target allocation

### LINKED ISSUE #73: unpriv-setup: make sure to exclude the caller's current uid/gid
If caller's uid or gid falls within its subuid or subgid allocation, then newuidmap/newgidmap will refuse to allow the mappings, causing stacker to fail.

tych0:
Closes #73

Signed-off-by: Tycho Andersen <tycho@tycho.ws>

### DIFF
--- a/cmd/unpriv-setup.go
+++ b/cmd/unpriv-setup.go
@@ -69,7 +69,7 @@ func warnAboutNewuidmap() {
 	}
 }
 
-func addSpecificEntries(file string, name string) error {
+func addSpecificEntries(file string, name string, currentId int) error {
 	content, err := ioutil.ReadFile(file)
 	if err != nil && !os.IsNotExist(err) {
 		return errors.Wrapf(err, "couldn't read %s", file)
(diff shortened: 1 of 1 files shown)


cmd/unpriv-setup.go
```diff
@@ -107,23 +107,32 @@ func addSpecificEntries(file string, name string) error {
 
 	}
 
+	// newuidmap (and thus lxc-usernsexec, or more generally liblxc) will
+	// complain if the current uid is in the subuid allocation. So if it
+	// is, let's just advance the subuid allocation another 65536 uids. we
+	// don't need to check if this overlaps again, since we know that
+	// maxAlloc was the highest existing allocation.
+	if maxAlloc < currentId && currentId < maxAlloc + 65536 {
```
hallyn: Does the first one need to be <= ? if currentId is 100000 and maxAlloc is 100000, we want it to skip.

tych0: Yes, yes it does. Thanks :)

hallyn:
Cool, thanks, I'd looked for a minute and wasn't sure the best way to do it, looks perfect :)

PR #75, opened 2020-02-15
Only print an error if there was an error

hallyn:
Fix an error in my last patch which causes us to always print something
like

2020/02/14 22:38:37  warn failed parsing /etc/sub{u,g}idmap: <nil>


tych0:
Oh, I think this is my fault actually. But yes, looks good :)

PR #76, opened 2020-02-15
upriv squashfs

hallyn:
(creating this PR now, but it's not yet ready)


rchamarthy:
Looks like there is a way to mark PR as draft so it’s not mergable but reviews etc can happen.

hallyn:
Patch should be ready now.

PR #78, opened 2020-03-05
Allow substitutions from stacker config into a stacker.yaml

smoser:
This allows the stacker.yaml to reference stacker config.
It allows substitutions of:

   STACKER_ROOTFS_DIR  RootFSDir
   STACKER_STACKER_DIR StackerDir
   STACKER_OCI_DIR     OCIDir


tych0:
Would be good to add a test to make sure this keeps working, but the patch itself looks fine, thanks!

PR #79, opened 2020-03-05
Add Config-value documentation and a test. Remove an inadvertant printf.

smoser:
Follow on to #78 to:
   - Remove debug Printf.
   - Add test for config substitutions.
   - Add some helpers for tests (umount_under, stderr, tmpd, cmp_files)
   - Add some doc for config file substitutions.

### DIFF
--- a/build.go
+++ b/build.go
@@ -195,7 +195,6 @@ func (b *Builder) Build(file string) error {
 		os.RemoveAll(opts.Config.StackerDir)
 	}
 
-	fmt.Printf("Build subs: %v\n", b.opts.Config.Substitutions())
 	sf, err := NewStackerfile(file, append(opts.Substitute, b.opts.Config.Substitutions()...))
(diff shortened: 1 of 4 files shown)


tych0:
Seems like the test is busted?

smoser:
Test runs for me locally... Give it a try?

tych0:
Too bad we don't run CI on your machine :)

I think it's because you're using a non-broken lxc-common package that has lxc.init.static. The ones in ubuntu are broken, and it's not statically linked, so it fails to start. When I switch to using centos:latest instead of busybox:latest, it passes: https://travis-ci.org/tych0/stacker/builds/658867521

smoser:
@tych0 
Take another look.  Moved back to centos :cry: . And re-worked some cleanup.

tych0:
Cool, thanks!

PR #80, opened 2020-03-06
doc: update install.md with F31 instructions

pcmoore:
In addition to the F31 instructions there are some layout tweaks to
better support instructions for more than one Linux distro.

Signed-off-by: Paul Moore <pmoore2@cisco.com>

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -1,15 +1,35 @@
-### Building
+## Building and Installing Stacker
 
-Stacker requires at least go 1.11, since it uses go modules.
-On Ubuntu 18.04 you can get that using the instructions at:
+### Go Dependency
+
+Stacker requires at least go 1.11.
+
+#### Ubuntu 18.04
+
+On Ubuntu 18.04 you can install Go using the instructions at:
 https://github.com/golang/go/wiki/Ubuntu
 
-Or you could get the latest version available at:
+#### Fedora 31
+
+On Fedora 31 you can install Go with the following command:
+
+    sudo dnf install golang
+
+#### Other Distributions
+
+If Go is not already packaged for your Linux distribution, you can get the
+latest Go version here:
 https://golang.org/dl/#stable
-And install it using the instructions on on the official go website:
+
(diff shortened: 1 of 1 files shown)


doc/install.md
```diff
@@ -1,15 +1,35 @@
-### Building
+## Building and Installing Stacker
 
-Stacker requires at least go 1.11, since it uses go modules.
-On Ubuntu 18.04 you can get that using the instructions at:
+### Go Dependency
+
+Stacker requires at least go 1.11, since it uses Go modules.
```
smoser: I'd just scratch ', since it uses Go modules'.

1.11 is 18 months old.

pcmoore: That was just a carryover from the original file, I'll drop it.

doc/install.md
```diff
@@ -1,15 +1,35 @@
...
+
+#### Ubuntu 18.04
+
+On Ubuntu 18.04 you can install Go using the instructions at:
 https://github.com/golang/go/wiki/Ubuntu
 
-Or you could get the latest version available at:
+#### Fedora 31
```
smoser: I dont' really think that it is of much use to tell people how to get golang. Our instructions are out of date, and incomplete. I'd just give a link to golang wiki and say figure it out.

pcmoore: Once again, carryover from the original docs; I don't really have much of a preference.  If @tych0 wants it gone, I'll be happy to drop it.

pcmoore:
> thanks for pull request. I don't want to sound negative... thank you.

No worries, I understand :)

> But I think most of this is probably "less is more", because "more" requires more maintenance.

Hey, I'm fine if you guys just want to drop this PR; I have no real attachment to it either way.  It only came about because I noticed that the package names weren't quite the same between Ubuntu and Fedora and I asked @tych0 if he was interested in a doc update PR; he said yes so here we are.  Feel free to do with it what you will :)

tych0:
LGTM, thanks!

PR #90, opened 2020-05-15
More detailed debug message in case layers are pruned from cache

andaaron:


tych0:
Arg. Looks like the build failing is my fault. I'll fix that, but since it built fine, this patch seems fine.

PR #92, opened 2020-05-21
Check roots dir exists before cleaning it

peusebiu:


cmd/clean.go
```diff
@@ -27,13 +27,16 @@ func doClean(ctx *cli.Context) error {
...
-		fail = true
+	if _, err := os.Stat(config.RootFSDir); !os.IsNotExist(err) {
+		// path/to/whatever exists
+		err := stacker.CleanRoots(config)
+		if err != nil {
+			log.Infof("problem cleaning roots, btrfs is probably wedged: %v", err)
+			fail = true
+		}
+		os.RemoveAll(config.RootFSDir)
+		os.RemoveAll(config.OCIDir)
```
tych0, line 38: Seems like these should still be outside of the test?

peusebiu: yes, you are right, I have updated it. thank you.

tych0:
Thanks!

PR #96, opened 2020-07-13
Add STACKER_LAYER_NAME to the name of the layer to a build environment.

### LINKED ISSUE #95: provide some way (environment?) to determine if inside a stacker
Our make system is aware of stacker, and will do some things (like write a stacker config) before invoking stacker.
The makefile gets invoked both inside and outside of stacker.

I'd like to have some logic that says "if you're already inside stacker, then you don't need to write a stacker config".

Can we provide some way to allow figure out if you're inside a stacker environment or not ?

maybe something like: STACKER_ENVIRONMENT=the_layer_name

I'd get that right now with the following, but i'd like to not have to define my own STACKER_ENVIRONMENT in the build_env.

```yaml
my_builder:
  from:
    type: docker
    ...
  build_env: {"STACKER_ENVIRONMENT": "my_builder"}
  run: |
    echo hi world. you are in ${STACKER_ENVIRONMENT}
  build_only: true

     

```

smoser:
Now a run section will contain an environment variable
STACKER_LAYER_NAME.  The value is the name of the layer section
in the yaml file.

Fixes #95.

### DIFF
--- a/types/layer.go
+++ b/types/layer.go
@@ -37,7 +37,6 @@ type Layer struct {
 	Binds              interface{}       `yaml:"binds"`
 	Apply              []string          `yaml:"apply"`
 	RuntimeUser        string            `yaml:"runtime_user"`
-	Name               string            // the name of this layer in the stacker file.
 	referenceDirectory string            // Location of the directory where the layer is defined
 }
 
@@ -93,9 +92,9 @@ func buildEnv(passThrough []string, newEnv map[string]string,
 	return ret, nil
 }
 
-func (l *Layer) BuildEnvironment() (map[string]string, error) {
+func (l *Layer) BuildEnvironment(name string) (map[string]string, error) {
 	env, err := buildEnv(l.BuildEnvPt, l.BuildEnv, os.Environ)
-	env["STACKER_LAYER_NAME"] = l.Name
+	env["STACKER_LAYER_NAME"] = name
 	return env, err
 }
 
--- a/types/layer.go
+++ b/types/layer.go
@@ -37,6 +37,7 @@ type Layer struct {
 	Binds              interface{}       `yaml:"binds"`
 	Apply              []string          `yaml:"apply"`
 	RuntimeUser        string            `yaml:"runtime_user"`
+	Name               string            // the name of this layer in the stacker file.
 	referenceDirectory string            // Location of the directory where the layer is defined
 }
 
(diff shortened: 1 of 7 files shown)


types/layer.go
```diff
@@ -37,6 +37,7 @@ type Layer struct {
 	Binds              interface{}       `yaml:"binds"`
 	Apply              []string          `yaml:"apply"`
 	RuntimeUser        string            `yaml:"runtime_user"`
+	Name               string            // the name of this layer in the stacker file.
```
tych0, line 40: Can we avoid doing this?

types/layer.go
```diff
@@ -93,7 +94,9 @@ func buildEnv(passThrough []string, newEnv map[string]string,
 }
 
 func (l *Layer) BuildEnvironment() (map[string]string, error) {
-	return buildEnv(l.BuildEnvPt, l.BuildEnv, os.Environ)
+	env, err := buildEnv(l.BuildEnvPt, l.BuildEnv, os.Environ)
+	env["STACKER_LAYER_NAME"] = l.Name
```
tych0, line 98: Is there a reason to have this separately from STACKER_STACKER_DIR and friends? I don't see any.

smoser: STACKER_STACKER_DIR are "stacker config"  level items.  this is a "layer" level item.  So substitution doesn't really make sense.

```yaml
my-build1:
  run: |
    this is ${STACKER_LAYER_NAME} of my-build1
my-build2:
  run: |
     here it would be my-build2
```

types/layer.go
smoser: well right now, given:

```yaml
foo:
  binds:
  apply:
  run: ...
```

We do not have a reference in the `Layer` to `foo` at all.
we could add a reference to the StackerConfig that this layer came from, but that seems more recursive than this.

types/layer.go
tych0: Ah, yep, makes sense.

types/layer.go
tych0: Why not just pass it as an argument to BuildEnvironment(), then?

smoser: it seems sane to me that a Layer has a name that was tied to it at yaml read time.
Do you have a reason not to?

tych0: Yes, you can then cache layer builds that were renamed; one of our products builds relied on this at one point, not sure if it still does or not.

smoser:
I just added a commit (56e13b07346e237f9f8c1db924e59c83428bb353) that changes this to pass the name into `BuildEnvironment` rather than having it on the Layer.

tych0:
Cool, thanks!

PR #103, opened 2020-10-27
Support use of squashtool for extraction.

smoser:
Unsquashfs just does not operate really well when it is not root.
It sometimes ignores extraction errors, and other times fails.
squashtool intends to behave better when unprivileged.

If unprivileged and user has squashtool in their path, then
squashtool will be used for extraction rather than unsquashfs.

By default:
  - as root, behave as before.
  - as non-root: use squashtool if available, else unsquashfs.

The decision of whether or not the user is operating as root can
be affected by STACKER_UNPRIV in the environment.
STACKER_UNPRIV can have three values:
 * "" (or not set in env): unprivileged unless euid == 0
 * true: unprivileged (non-root)
 * false: privileged (root)

### DIFF
--- a/cmd/umoci.go
+++ b/cmd/umoci.go
@@ -28,6 +28,7 @@ import (
 	"github.com/opencontainers/umoci/pkg/mtreefilter"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
+	"golang.org/x/sys/unix"
 )
 
 var umociCmd = cli.Command{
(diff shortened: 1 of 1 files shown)


cmd/umoci.go
```diff
@@ -188,12 +189,79 @@ func tarUnpack(oci casext.Engine, tag string, bundlePath string, callback layer.
...
+	err := os.MkdirAll(extractDir, 0755)
+	if err != nil {
+		return err
+	}
+
+	var uCmd []string
+	if which("squashtool") != "" {
+		uCmd = []string{"squashtool", "extract", "--whiteouts", "--perms"}
+		if !rootless {
+			uCmd = append(uCmd, "--devs", "--sockets", "--owners")
```
tych0, line 231: Doesn't this introduce correctness issues around permissions and such?

smoser:
For context, the reason I needed this was trying to build the following stacker.yaml as non-root.

```yaml
kernel-update-boot:
  from:
    type: oci
    url: oci:boot
  import:
    - https://example.com/..../kernel.rpm
  run: |
    touch VERSION3
    rpm -q kernel
    rpm -Uvh --force /stacker/*.rpm
    rpm -q kernel
    echo v3 > /version
```

the referenced 'oci' is an ocirepo that contains squashfs items.

When using unsquashfs, stacker will ultimately die with:

    create_inode: failed to create character device /.../roots/kernel-update-boot/rootfs/var/cache/yum/x86_64/7.7.1908/base/primary.xml, because Operation not permitted

The 'primary.xml' file is a whiteout file (char device 0:0) and unpriv user
does nto have the ability to create character devices.

smoser:
More context... to make the stacker.yaml listed above completely work, I had to fix an issue in squashtool (https://github.com/anuvu/squashfs/pull/5) as it would previously not extract /bin/sh with execute perms unless acting as root.  that is fixed now.

tych0:
For whiteouts specifically, kernels newer than 5.8 will allow unprivileged whiteout creation: https://github.com/anuvu/stacker/blob/master/overlay/overlay.go#L81

cmd/umoci.go
tych0: It seems like we should probably run unsquashfs in a user+mount+net ns instead like we do with umoci, so it can write all of these kinds of things.

smoser: It will not retain owners correctly. but i'm not sure how you could hope to do that as non-root.  perhaps if the process was executed in the user-namespace, and it used `--whiteouts --perms --owners` then it could do the right thing.

smoser: I accept that this isn't perfect, but I'd suggest that if you want to ensure correctness, then the right thing to do is just fail, as unsquashfs can't offer correctness either.

smoser: > It seems like we should probably run unsquashfs in a user+mount+net ns instead like we do with umoci, so it can write all of these kinds of things.

the issue with unsquashfs is that it does not understand whiteouts. Without explicit knowledge of them, you're destined to fail.  As you'll end up doing:

   * extract a layer with /some/dir
   * extract a layer with /some/dir/.whiteout

and unsquashfs will try to create the .whiteout file, which is a char device over the top of the existing 'dir' and will fail with "is a directory".

smoser: so i guess there are 2 use cases, and I had only considered one.

 1. extract layer1, layer2, layer3 to extractDir/

    here the goal is to have extractDir/ have the same contents that overlay
    mounts would give you of layer3 over layer2 over layer1.


 2. extract layer1 to extractDir
    here the goal is to have exactly the contents of layer1


The difference really is that you want '--whiteouts' on 1 but not on '2'.  You have to know which one you're doing.

I had only considered '1'.

tych0:
I'll close this, as a version of this patch landed in #104

PR #104, opened 2020-10-29
Err on no squashtool btrfs

tych0:
A superset of #103; this makes it an error not to have squashtool when using squashfs imports with the btrfs backend.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,5 +17,8 @@ before_install:
     - GO111MODULE=off go get github.com/opencontainers/umoci/cmd/umoci
     - sudo cp $GOPATH/bin/umoci /usr/bin
     - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.24.0
+    - sudo apt-get install -yy autoconf automake make autogen autoconf libtool binutils git squashfs-tools
+    - (cd /tmp && git clone https://github.com/AgentD/squashfs-tools-ng && cd squashfs-tools-ng && ./autogen.sh && ./configure --prefix=/usr && make -j2 && sudo make -j2 install && sudo ldconfig -v)
+    - (cd /tmp && git clone https://github.com/anuvu/squashfs && cd squashfs && make && sudo cp squashtool/squashtool /usr/bin)
 script:
     - env GO111MODULE=on BATS=/usr/bin/bats make check
--- a/cmd/umoci.go
+++ b/cmd/umoci.go
@@ -28,6 +28,7 @@ import (
 	"github.com/opencontainers/umoci/pkg/mtreefilter"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
+	"golang.org/x/sys/unix"
 )
 
 var umociCmd = cli.Command{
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -17,5 +17,8 @@ before_install:
     - GO111MODULE=off go get github.com/opencontainers/umoci/cmd/umoci
     - sudo cp $GOPATH/bin/umoci /usr/bin
     - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.24.0
+    - sudo apt-get install -yy autoconf automake make autogen autoconf libtool binutils git squashfs-tools
+    - (cd /tmp && git clone https://github.com/AgentD/squashfs-tools-ng && cd squashfs-tools-ng && ./autogen.sh && ./configure --prefix=/usr && make -j2 && sudo make -j2 install && sudo ldconfig -v)
```
smoser, line 21: squashfs has code to do this (to get squashfs-tools-ng and build it).

As root, you should be able to:

    git clone https://github.com/anuvu/squashfs && cd squashfs && 
        ./setup --install-packages /usr/local &&
        make && cp squashtool/squashtool /usr/local/

squashfs obviously has the need to maintain that path as working.  So that is one less thing for you to have to worry about.

smoser: bah... to run the squashtool/squashtool you also need '/sbin/ldconfig'.
For your purposes, maybe 

    make static && cp squashtool/squashtool.static /usr/bin/squashtool

cmd/umoci.go
```diff
@@ -188,6 +189,61 @@ func tarUnpack(oci casext.Engine, tag string, bundlePath string, callback layer.
...
+}
+
+func extractSingleSquash(squashFile string, extractDir string) error {
+	err := os.MkdirAll(extractDir, 0755)
+	if err != nil {
+		return err
+	}
+
+	var uCmd []string
+	if config.StorageType == "btrfs" {
```
smoser, line 228: the decision on which tool to use isn't really based on the storageType, but rather the intent of the extract.

Either:
  a.) you are extracting as an archive (you want the .whiteout files present) so that you can overlay mount or maybe modify and re-pack
  b.) you are extracting to use the directory as if it is a layer.

Since we are assuming priviledged, the only difference is '--whiteouts'.

And just to be clear, `unsquashfs` can only be used for 'a' and requires that:
 1. you are privileged and can create device nodes (or squashFile has no device nodes)
 2. `extractDir` is empty or `squashFile` has no whiteout files.

cmd/umoci.go
```diff
@@ -188,6 +189,61 @@ func tarUnpack(oci casext.Engine, tag string, bundlePath string, callback layer.
...
+		if which("squashtool") == "" {
+			return errors.Errorf("must have squashtool to correctly extract squashfs using btrfs storage backend")
```
smoser, line 230: probably a link to how the user could obtain squashtool would be good. its not a household name yet.

cmd/umoci.go
tych0: But the "intent of the extract" here is exactly storage type: in btrfs we want it to unlink(), and in overlay we want it to mknod.

smoser: Is that true always?  i was thinking there was different intent between stacker using a layer (and putting it in ~/.stacker or wherever) and a stacker file doing the following, which was what caused me to open #103.

    from:
      type: oci
      url: oci:boot

.travis.yml
smoser: thats fine... and fwiw, it wasn't clear above, but `--install-packages` would handle your 'apt-get' line too.

cmd/umoci.go
tych0: Yes, it is always true. When you see from: type: oci, stacker still extracts the layer using the storage backend so you can run code on it.

tych0: So there was some code a while ago that probably handled this extraction from tar files incorrectly, https://github.com/opencontainers/umoci/commit/c0f84452c755c30beb2d199121fd096e95656299 and https://github.com/opencontainers/umoci/commit/a425f92e910345405e8c8d8bb67e7da510ac1b65 would have fixed some of that, though. Maybe that's what you were seeing?

smoser: shrug.  I dont' know.  I thought it was using unsquashfs to extract the oci:boot layer (including the underlying layers) to do the import.  That failed because unsquashfs can't handle .whiteout files.

if its fixed, then good.  If its not, then i'll bother you again sometime.
I suspect my  usecase will be fixed by  use of squashtool here.

PR #107, opened 2020-11-23
lib/image: added SkipTLS option for source and destination separately

shimish2:
1. Add separate options for source and destination skip tls verification. 
2. Since SkipTLS verification was limited to source before, replaced SkipTLS option with SrcSkipTLS option wherever SkipTLS option is set.


tych0:
Thanks!

Issue #113, opened 2020-12-18
grab: don't exec cp in the layer


tych0:
This assumes that the layer has cp, which some layers don't.

We could bind mount ourselves in and have a `cp` thing which calls `shutil.CopyTree()`, but it really feels like we should just fix #111 first, and then we can just use a `shutil.CopyTree` here instead of having to use a container at all.

PR #114, opened 2021-01-06
Copy using CopyDirectory rather than exec to cp.

smoser:
Trying to use cp after being in a namespace then depends
on access to that executable.


smoser:
fixes an issue for me, related to #113

tych0:
Seems like this breaks in the face of broken symlinks in the input dir?

tych0:
https://github.com/otiai10/copy looks a little more active, and has options to specify what happens on symlinks. I'm going to take a whack at fixing #113 and see how it works.

PR #132, opened 2021-02-02
Support speciying hash of import source, closes #119

### LINKED ISSUE #119: allow specifying hash of import source
for import entries, allow specifying the hash so that stacker can check that the bits you get are the bits you want

something like this

```yaml
first:
    from:
        type: docker
        url: docker://centos:latest
    import:
        - path: "config.json"
          hash: "BEEFcafeaaaaAAAA...."
```

RFC: maybe also add a flag to have stacker reject any stackerfiles that have imports with no hash?

peusebiu:
- supports both imports style:
```
import:
  - path: path/to/file
    hash: sha256sum
  - stacker://first/example.txt
  - http://url/to/file
```

### DIFF
--- a/doc/stacker_yaml.md
+++ b/doc/stacker_yaml.md
@@ -84,6 +84,31 @@ will not be reflected.
 
 Will grab /path/to/file from the previously built layer `$name`.
 
+#### `import hash`
+
+The `import` directive also supports specifying the hash(sha256sum) of import source,
+for all the three forms presented above, for example:
+```
+import:
+  - path: config.json
+    hash: f55af805b012017bc....
+  - path: http://example.com/foo.tar.gz
+    hash: b458dfd63e7883a64....
+  - path: stacker://$name/path/to/file
+    hash: f805b012017bc769a....
+```
+Before copying the file it will check if the requested hash matches the actual one.
+This new import mode can be combined with the old one, for example:
+```
+import:
+  - path: "config.json
+    hash: "BEEFcafeaaaaAAAA...."
+  - /path/to/file
+```
+
+
+
+
 #### `environment`, `labels`, `working_dir`, `volumes`, `cmd`, `entrypoint`, `user`
 
 These all correspond exactly to the similarly named bits in the [OCI image
--- a/import.go
+++ b/import.go
@@ -1,10 +1,12 @@
 package stacker
 
 import (
+	"github.com/opencontainers/go-digest"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path"
+	"strings"
 
 	"github.com/anuvu/stacker/lib"
 	"github.com/anuvu/stacker/log"
@@ -82,13 +84,35 @@ func chmodParentAndRemove(destpath string) (func(), error) {
 	return func() { os.Chmod(dir, orig.Mode()) }, os.RemoveAll(destpath)
 }
 
-func importFile(imp string, cacheDir string) (string, error) {
+func verifyImportFileHash(imp string, hash string) error {
+	if len(hash) == 0 {
+		return nil
+	}
+	actualHash, err := lib.HashFile(imp, false)
+	if err != nil {
+		return err
+	}
+
+	actualHash = strings.TrimPrefix(actualHash, "sha256:")
+	if actualHash != strings.ToLower(hash) {
+		return errors.Errorf("The requested hash of %s import is different than the actual hash: %s != %s",
+			imp, hash, actualHash)
+	}
+
+	return nil
+}
+
+func importFile(imp string, cacheDir string, hash string) (string, error) {
 	e1, err := os.Lstat(imp)
 	if err != nil {
 		return "", errors.Wrapf(err, "couldn't stat import %s", imp)
 	}
 
 	if !e1.IsDir() {
+		err := verifyImportFileHash(imp, hash)
+		if err != nil {
+			return "", err
+		}
 		needsCopy := false
 		dest := path.Join(cacheDir, path.Base(imp))
 		e2, err := os.Stat(dest)
@@ -99,7 +123,6 @@ func importFile(imp string, cacheDir string) (string, error) {
 			if err != nil {
 				return "", err
 			}
-
 			needsCopy = differ
 		}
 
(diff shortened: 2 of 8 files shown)


grab.go
```diff
@@ -21,5 +22,12 @@ func Grab(sc types.StackerConfig, storage types.Storage, name string, source str
 	}
 	defer os.Remove(path.Join(sc.RootFSDir, name, "rootfs", "stacker"))
 
+	if len(hash) > 0 {
+		if err = c.Execute(fmt.Sprintf("echo %s %s | sha256sum --check", hash, source), nil); err != nil {
```
peusebiu, line 26: This seems to not work properly. Am I missing something here?

tych0: Yes -- Execute() just uses execv() ultimately, and thus these things are not running inside of a shell, so using shell syntax like `|` here will not work.

Additionally, it would be good if we can avoid executing stuff inside the container if at all possible. Can we hash this on the host after we've imported it? I believe this will keep all the import hashing code in one place as well.

peusebiu: Understood, I will hash it on the host. Thank you for the detailed answer.

import.go
```diff
@@ -256,7 +278,7 @@ func CleanImportsDir(c types.StackerConfig, name string, imports []string, cache
 	return nil
 }
 
-func Import(c types.StackerConfig, storage types.Storage, name string, imports []string, progress bool) error {
+func Import(c types.StackerConfig, storage types.Storage, name string, imports []map[string]string, progress bool) error {
```
tych0, line 281: Can we make this a new type, i.e. `[]Import` or similar instead of just a map?

types/layer.go
```diff
@@ -125,19 +126,61 @@ func (l *Layer) ParseFullCommand() ([]string, error) {
...
-	rawImports, err := l.getStringOrStringSlice(l.Import, func(s string) ([]string, error) {
-		return strings.Split(s, "\n"), nil
-	})
-	if err != nil {
-		return nil, err
+// Import directive from stacker file needs to support both []string and []map[string]string eg:
+// Layer.Import: [{"hash": importHash, "path": importPath}] && [importPath1, importPath2]
+// UpdateLayerImports converts Layer.Import from []interface{} to []map[string]string
+// Make sure we call this before reading stacker files and before opening cache
+func (l *Layer) UpdateLayerImports() error {
```
tych0, line 133: Instead of mutating `l` here, can we just have this return the list of parsed imports?

peusebiu:
Hello @tych0 Thanks for the review, I was going to ping you here.

Yes, I figured it out that handling maps all over the code base would be a pain, so I refactored it and now I'm using a slice of struct for Imports. And all the type casting is done in UnmarshalJSON and UnmarshalYAML interfaces.

types/layer.go
```diff
@@ -49,6 +59,64 @@ type Layer struct {
...
+
+func customUnmarshal(im *ImportMaps, data interface{}) {
+	imports, ok := data.([]interface{})
+	if ok {
+		// imports are a list of either strings or maps
+		for _, v := range imports {
+			*im = append(*im, getImportMapFromInterface(v))
+		}
+	} else {
+		log.Debugf("data: %#v\n", data)
```
peusebiu, line 90: Forgot it, will remove it after the build finishes

network.go
```diff
@@ -16,10 +16,27 @@ import (
...
 	name := path.Join(cacheDir, path.Base(url))
 
+	remoteHash, remoteSize, err := getHttpFileInfo(url)
+	if err != nil {
+		// Needed for "working offline"
+		// See https://github.com/anuvu/stacker/issues/44
+		log.Infof("cannot obtain file info of %s", url)
+	}
+	log.Debugf("Remote file: hash: %s length: %s", remoteHash, remoteSize)
+	if hash != "" && hash != remoteHash {
```
peusebiu, line 29: Should I move this hash check into import.go?

types/layer.go
```diff
@@ -29,9 +32,16 @@ func IsContainersImageLayer(from string) bool {
 	return false
 }
 
+type ImportMap struct {
```
tych0, line 35: I think you can drop the "Map" suffix here, and just call it "Import"

network.go
tych0: Yeah, that might be nice if we can figure out how to do it. Then we'd only have to hash the file once if both hashes were specified.

tych0:
Can you also add a test for this functionality?

peusebiu:
> Can you also add a test for this functionality?

I've added one.

test/import.bats
```diff
@@ -72,3 +72,36 @@ EOF
...
+    test_file_sha=$(sha test_file) || { stderr "failed sha $test_file"; return 1; }
+    touch test_file2
+    cat > stacker.yaml <<EOF
+first:
+    from:
+        type: oci
+        url: $CENTOS_OCI
+    import:
+        - path: test_file
+          hash: $test_file_sha
```
tych0, line 87: Can we test the failure case too, where the sha doesn't match? There should be a `stacker_bad` function you can use to expect failure of stacker.

peusebiu: Ok, sure, thank you.

doc/stacker_yaml.md
```diff
@@ -84,6 +84,31 @@ will not be reflected.
...
+    hash: b458dfd63e7883a64....
+  - path: stacker://$name/path/to/file
+    hash: f805b012017bc769a....
+```
+Before copying the file it will check if the requested hash matches the actual one.
+This new import mode can be combined with the old one, for example:
+```
+import:
+  - path: "config.json
+    hash: "BEEFcafeaaaaAAAA...."
```
tych0, line 105: You give what looks to be a case insensitive example here, but as far as I can see the actual comparison will be case sensitive. It might be nice to make this case insensitive, and also give an error if "hash" isn't an actual valid hash.

Other than that, this looks good to me, thanks!

peusebiu: Ok, I pushed the update making the hash insensitive and also checking for invalids hashes.

tych0:
Great, thanks for working on this!

PR #134, opened 2021-02-16
ci: fix skopeo install

tych0:
The skopeo install is failing with:

Unpacking containers-common (100:1-7) over (1.2.1~2) ...
dpkg: error processing archive /tmp/apt-dpkg-install-at4ZRl/03-containers-common_100%3a1-7_all.deb (--unpack):
 trying to overwrite '/usr/share/man/man5/containers-auth.json.5.gz', which is also in package containers-image 5.8.1~1

we get our skopeo install from the opensuse build infra for ubuntu, which
is presumably not well maintained :)

In 20.10 it is included in the main ubuntu repos, so let's just get it from
there.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,7 +8,7 @@ on:
 
 jobs:
     build:
-        runs-on: ubuntu-20.04
+        runs-on: ubuntu-20.10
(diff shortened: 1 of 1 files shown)


tych0:
github actions doesn't support 20.10, so this won't work :(

PR #137, opened 2021-03-04
Reexecute stacker build inside lxc-usernsexec, do not reexecute inter…

### LINKED ISSUE #111: Make things function better in user namespaces
Due to the limitations in go, right now if we want to run stuff in a user namespace, we have to re-exec things.

When stacker first started out, this was really limited to umoci extraction, but as it has gained more and more features `stacker://` style imports, etc, more and more regular filesystem operations will just fail when interacting with operations that had previously been done in a user namespace.

Instead, perhaps we should just run everything in a user namespace, and the parent stacker process would just set up the user namespace and then re-exec stacker with the right args. This would let us get rid of a lot of the ugly re-exec points.

peusebiu:
…nal go commands, closes #111


cmd/build.go
```diff
@@ -99,7 +101,74 @@ func doBuild(ctx *cli.Context) error {
 	if err != nil {
 		return err
 	}
+	if !args.Config.Userns {
+		binary, err := os.Readlink("/proc/self/exe")
+		if err != nil {
+			return err
+		}
 
+		cmd := []string{
```
tych0, line 110: Here instead of duplicating all the arguments, can we replace os.Args[0] and add `--internal-userns` as Args[1] instead? This way when people add new arguments they won't also have to change this code.

cmd/build.go
```diff
@@ -99,7 +101,74 @@ func doBuild(ctx *cli.Context) error {
...
+		if len(ctx.String("stacker-file")) > 0 {
+			cmd = append(cmd, "-f")
+			cmd = append(cmd, ctx.String("stacker-file"))
+		}
+
+		err = container.MaybeRunInUserns(cmd, "Build in container")
+		return err
+	}
 	builder := stacker.NewBuilder(&args)
 	return builder.BuildMultiple([]string{ctx.String("stacker-file")})
```
tych0, line 173: How does this work if someone invokes recursive-build?

container/container.go
```diff
@@ -103,6 +102,14 @@ func RunInUserns(idmapSet *idmap.IdmapSet, userCmd []string, msg string) error {
 	return nil
 }
 
+func Run(userCmd []string, msg string) error {
```
tych0, line 105: Is this used anywhere?

tych0:
Overall this looks great, thanks for working on it. If nothing else, it lets us get rid of 150 lines of terrible re-exec code!

A couple of small nits above and unfortunately, I added another re-exec command today, so this will need a rebase to account for that too.

Finally, it seems like we could split this into two commits: 1. adding the global re-exec and 2. removing the use of internal_go.go which may be easier for future readers/reviewers.

cmd/build.go
peusebiu: Sure, haven't thought about this way. Thanks!

cmd/build.go
peusebiu: Yes, I forgot about recursive builds. Thank you!

container/container.go
peusebiu: Will remove it, thanks

peusebiu:
Thanks for the review, I have updated it according to your review, I hope I understand correctly the split commit part, if not, I can update it again.

Also, I'm not sure if I have to remove the whole internal-go.go file, or we leave the test commands there, like I did.

Thanks again,
Sebi

tych0:
The code itself looks good, but I'm not seeing it actually use the --internal-userns flag anywhere as far as I can tell:

```
tycho    3527831  0.0  0.0  14980 10928 pts/3    Ss   09:51   0:00          \_ -bash
tycho    3528576 13.4  0.0 1392936 22688 pts/3   Sl+  09:58   0:00              \_ stacker --storage-type=overlay build
tycho    3528641  0.0  0.0  24576  5180 ?        Ss   09:58   0:00                  \_ /home/tycho/stacker/stacker internal test /tmp/foo/roots /tmp/stacker_test_run581706474
tycho    3528642  0.0  0.0   1340     4 ?        Ss   09:58   0:00                      \_ init
tycho    3528647  0.0  0.0  11896  2852 ?        Ss   09:58   0:00                          \_ /bin/sh -xe /stacker/.stacker-run.sh
tycho    3528648  0.0  0.0  23028  1408 ?        S    09:58   0:00                              \_ /usr/bin/coreutils --coreutils-prog-shebang=sleep /usr/bin/sleep 200s
```

with the stacker file:

```
test:
    from:
        type: oci
        url: base:centos
    run: |
        sleep 200s
```

I would expect to see another subprocess in the middle there with an `--internal-userns` argument somewhere?

It might be useful to add the testcase from https://github.com/anuvu/stacker/issues/116#issuecomment-760470939 in another commit in this series, I think that will ensure that things work correctly.

tych0:
One other nit: the subject line of the commits is pretty long (and thus rendered strangely by github); I find it is generally worth following the 50 character limit (with a 72 char limit for the message body). More details here: https://chris.beams.io/posts/git-commit/

peusebiu:
Hello, this is my output, I can't understand why it behaves like this on your machine:

```
 2000 pts/0    Ss     0:00  |   \_ bash
 2554 pts/0    Sl+    0:00  |   |   \_ ./stacker --debug build -f tycho.yaml
 2559 pts/0    S+     0:00  |   |       \_ lxc-usernsexec -m u:0:1000:1 -m u:1:100001:65535 -
 2560 pts/0    Sl+    0:04  |   |           \_ /home/peusebiu/unpriv/stacker --internal-usern
 2577 ?        Ss     0:00  |   |               \_ /home/peusebiu/unpriv/stacker internal tes
 2580 ?        Ss     0:00  |   |                   \_ init
 2581 ?        Ss     0:00  |   |                       \_ /bin/sh -xe /stacker/.stacker-run.
 2582 ?        S      0:00  |   |                           \_ /usr/bin/coreutils --coreutils
```

I will add the test and modify the comment. Thanks.

peusebiu:
Can you please run @test "clean of unpriv overlay works", because I don't have a >5.8 kernel, and I'm not sure it ran in the CI:
```
ok 24 clean of unpriv overlay works # skip need newer kernel for unpriv overlay
```

tych0:
> Yeah, I think the check for overlay kernels is actually broken by this patch somehow.

Looks like it's this hunk:

```
@@ -193,7 +50,7 @@ func doTestsuiteCheckOverlay(ctx *cli.Context) error {
                return errors.Wrapf(err, "couldn't make rootfs dir for testsuite check")
        }
 
-       err = container.RunInternalGoSubcommand(config, []string{"check-overlay"})
+       err = overlay.CanDoOverlay(config)
        if err != nil {
                log.Infof("%s", err)
                os.Exit(50)
```

This test still needs to be run inside a userns, as only root in a userns owning the mountns can do a `mount()`.

tych0:
Ok, I pushed the fix for the testsuite-check-overlay command (and also dropped some unused stuff from the first patch). I also dropped the clean test from the series, since it doesn't actually fix things right now. I'm working on that now, though.

tych0:
Ok, I think with this last commit this is ready to go. @peusebiu thoughts? Assuming it looks good to you, we can go ahead and merge and see what happens :)

peusebiu:
Nice, I missed some obvious things, thanks for patching them :D.
Looks good to me. Let's see, hope for the best.

tych0:
No problem, thanks for your work on this!

PR #140, opened 2021-03-10
cache: bump cache version + add a test

### LINKED ISSUE #119: allow specifying hash of import source

tych0:
If we run this new test with:

diff --git a/test/caching.bats b/test/caching.bats
index b2091d2..bbb0b20 100644
--- a/test/caching.bats
+++ b/test/caching.bats
@@ -185,7 +185,7 @@ EOF

     git clone https://github.com/anuvu/stacker
-    (cd stacker && make)
+    (cd stacker && git co 481c315c8ca3c930fb8a215df8baf0b06bc38bc7 && make)

     # note that this is not quite a complete test: since the cache has a
     # types.Layer in it, anything that can change in there could break this. so

we get:

error: json: cannot unmarshal string into Go struct field Layer.cache.Layer.Import of type types.Import

This is because the commit after 481c315, 60cb730340ac ("Support
specifying hash of import source, closes #119") changed the cache
structure. We have a provision when the cache structure is changed: bump
the cache version so that things are completely invalidated and totally
rebuilt. However, this change snuck through without getting a cache version
bump. So, let's bump it now.

Let's also add a test so hopefully we can catch such things in the future. (As
noted in the test, it's not quite complete, but having some coverage is better
than none.)

### DIFF
--- a/test/caching.bats
+++ b/test/caching.bats
@@ -182,3 +182,25 @@ EOF
     umoci unpack --image oci:mode-test dest
     [ -x dest/rootfs/executable ]
 }
+
+@test "can read previous version's cache" {
+    git clone https://github.com/anuvu/stacker
+    (cd stacker && make)
+
+    # some additional testing that the cache can be read by older versions of
+    # stacker (cache_test.go has the full test for the type, this just checks
+    # the mechanics of filepaths and such)
+    touch foo
+    cat > stacker.yaml <<EOF
+test:
+    from:
+        type: oci
+        url: $CENTOS_OCI
+    import:
+        - foo
+    run: cp /stacker/foo /foo
+EOF
+
+    ./stacker/stacker --storage-type=$STORAGE_TYPE build
+    stacker build
+}
--- a/test/caching.bats
+++ b/test/caching.bats
@@ -204,3 +204,15 @@ EOF
     ./stacker/stacker --storage-type=$STORAGE_TYPE build
     stacker build
 }
+
+@test "different old cache version is ok" {
+    cat > stacker.yaml <<EOF
+test:
+    from:
+        type: oci
+        url: $CENTOS_OCI
+EOF
+    stacker build
+    echo '{"version": 1, "cache": "lolnope"}' > .stacker/build.cache
+    stacker build
+}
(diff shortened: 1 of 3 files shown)


test/caching.bats
```diff
@@ -182,3 +182,25 @@ EOF
...
     [ -x dest/rootfs/executable ]
 }
+
+@test "can read previous version's cache" {
+    git clone https://github.com/anuvu/stacker
+    (cd stacker && make)
+
+    # note that this is not quite a complete test: since the cache has a
+    # types.Layer in it, anything that can change in there could break this. so
+    # really, any time we update that structure, we should update this test.
```
raharper, line 192: Could you expand on this point about updating this test? What would need updating?
It's also not obvious (to me) how this tests the cache layer format?  Without digging into
the harness, how do we know what cache dir the stacker command is using and what
version of the cache is in-use?

tych0: > Could you expand on this point about updating this test? What would need updating?

Sure, mostly it's that if you add a new attribute to layer, you should add it here so that we test that we can read+write it across cache versions. I should probably add something for everything that exists in layer right now, although I was trying to cook up a way to reflectively add these things in a test so it wouldn't be so manual.

> It's also not obvious (to me) how this tests the cache layer format? Without digging into
the harness, how do we know what cache dir the stacker command is using and what
version of the cache is in-use?

The cache version is written into the cache file itself by stacker. Since the test builds the github.com/anuvu/stacker master branch, it will always be the "old" cache version if the patch in question changes the cache version. The subsequent build is with the "new" version on whatever branch the test is being run, so if the cache can't be parsed by the new code, things will fail (modulo the point above about being incomplete).

Each test is run in its own subdirectory, so the cache dir is just the default .stacker for both invocations.

tych0: Ok, I added a reflection-based test that doesn't need all the caveats about "you have to remember to update this" to try and catch this problem in the future. I left this test here because it doesn't really hurt anything, but I dropped the comment about needing to update it.

PR #152, opened 2021-04-05
Copy imports into layer's rootfs

### LINKED ISSUE #83: add an overlay_dirs: yaml entry
There's a lot of patterns of:

```
import: foo
run: cp /stacker/foo /
```

or worse

```
import: foo
rsync --lots-of-rsync-options /stacker/foo /
```

If we could instead have stacker do that, it would remove the need for people to have rsync or cp in their filesystem, and also avoid them having to think about all the right rsync options to pass in, since we could just import things with the right permissions.

Indeed, overlay_dirs: could just be layers themselves that stacker would auto-generate before doing the build (still exposing them in the rfs, of course).

peusebiu:
Copy them into the specified destination
otherwise copy them into layer's rootfs "/"
Preserves links, executability, times.
Don't preserve user/grop ownership and perms.
closes #83

### DIFF
--- a/lib/file.go
+++ b/lib/file.go
@@ -31,17 +31,34 @@ func FileCopy(dest string, source string) error {
 	}
 	defer s.Close()
 
-	fi, err := s.Stat()
-	if err != nil {
-		return err
-	}
-
 	d, err := os.Create(dest)
 	if err != nil {
 		return err
 	}
 	defer d.Close()
 
+	fi, err := s.Stat()
+	if err != nil {
+		return err
+	}
+
+	//if preservePermissions {
+	//	err = d.Chmod(fi.Mode())
+	//	if err != nil {
+	//		return err
+	//	}
+	//} else {
+	//	// preserve executable
+	//	if IsExecOwner(fi.Mode()) {
+	//		err = d.Chmod(0700)
+	//		if err != nil {
+	//			return err
+	//		}
+	//	}
+	//	// preserve time
+	//	if err = os.Chtimes(dest, fi.ModTime(), fi.ModTime()); err != nil {
+	//		return err
+	//	}
 	err = d.Chmod(fi.Mode())
 	if err != nil {
 		return err
@@ -80,4 +97,8 @@ func FindFiles(base, pattern string) ([]string, error) {
 	err = filepath.Walk(base, visit)
 
 	return paths, err
+}
+
+func IsExecOwner(mode os.FileMode) bool {
+	return mode&0100 != 0
 }
--- a/build.go
+++ b/build.go
@@ -405,6 +405,13 @@ func (b *Builder) Build(s types.Storage, file string) error {
 			return err
 		}
 
+		// code for copy imports in rootfs
+		log.Debugf("Copying imports into %s's rootfs", name)
+		err = copyImportsInRootfs(name, imports, opts.Config, s)
+		if err != nil {
+			return err
+		}
+
 		c, err := NewContainer(opts.Config, s, name)
 		if err != nil {
 			return err
--- a/import.go
+++ b/import.go
@@ -1,6 +1,7 @@
 package stacker
 
 import (
+	"fmt"
 	"github.com/opencontainers/go-digest"
 	"io/ioutil"
 	"os"
@@ -279,7 +280,6 @@ func acquireUrl(c types.StackerConfig, storage types.Storage, i string, cache st
 		if err != nil {
 			return "", err
 		}
-
 		err = verifyImportFileHash(p, hash)
 		if err != nil {
 			return "", err
@@ -353,5 +353,77 @@ func Import(c types.StackerConfig, storage types.Storage, name string, imports t
 		}
 	}
 
+	return nil
+}
+
+// Copy imports to a container rootfs
+func copyImportsInRootfs(name string, imports types.Imports, sc types.StackerConfig, storage types.Storage) error {
+	cacheDir := path.Join(sc.StackerDir, "imports", name)
+	for _, i := range imports {
+		url, err := types.NewDockerishUrl(i.Path)
+		if err != nil {
+			return err
+		}
+		rootfsDir := "rootfs"
+		source := path.Join(cacheDir, path.Base(i.Path))
+
+		// logic for overlay storage
+		// we need to mount host's import dir
+		if storage.Name() == "overlay" {
+			c, err := NewContainer(sc, storage, name)
+			if err != nil {
+				return err
+			}
+			defer c.Close()
+
+			err = c.bindMount(cacheDir, "/stacker", "")
+			if err != nil {
+				return err
+			}
+			//defer os.Remove(path.Join(sc.RootFSDir, name, "rootfs", "stacker"))
+			// mkdir destination if it doesn't exist
+			dest := path.Join(i.Dest, path.Base(i.Path))
+			if i.Dest != "/" {
+				err = c.Execute(fmt.Sprintf("mkdir -p %s", i.Dest), nil)
+				if err != nil {
+					return err
+				}
+			}
+			err = c.Execute(fmt.Sprintf("cp -r --preserve=all --no-preserve=ownership /stacker/%s %s", path.Base(i.Path), dest), nil)
+			if err != nil {
+				return err
+			}
+			continue
+		}
+		// logic for btrfs storage
+		dest := path.Join(sc.RootFSDir, name, rootfsDir, i.Dest, path.Base(i.Path))
+		// mkdir destination if it doesn't exist
+		if _, err := os.Stat(dest); os.IsNotExist(err) {
+			if err = os.MkdirAll(dest, 0755); err != nil {
+				return err
+			}
+		}
+
+		if url.Scheme == "http" || url.Scheme == "https" {
+			// import is file
+			if err = lib.FileCopy(dest, source); err != nil {
+				return err
+			}
+		} else {
+			e1, err := os.Lstat(source)
+			if err != nil {
+				return err
+			}
+			if e1.IsDir() {
+				if err = lib.DirCopy(dest, source); err != nil {
+					return err
+				}
+			} else {
+				if err = lib.FileCopy(dest, source); err != nil {
+					return err
+				}
+			}
+		}
+	}
 	return nil
 }
--- a/lib/dir.go
+++ b/lib/dir.go
@@ -1,0 +1,67 @@
+package lib
+
+import (
+	"io/ioutil"
+	"os"
+	"path"
+)
+
+// DirCopy copies a whole directory recursively
+func DirCopy(dest string, source string) error {
+
+	var err error
+	var fds []os.FileInfo
+	var srcinfo os.FileInfo
+
+	if srcinfo, err = os.Stat(source); err != nil {
+		return err
+	}
+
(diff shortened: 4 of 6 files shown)


peusebiu:
I'm not 100% sure about what I did with permissions. I have some commented code, if you want to take a look.
Probably this can be improved.

Otherwise. The code do the following:
- by default copy all imports from stackerdir to container's rootfs: /
- import have a new directive "dest" (besides path and hash) which we can use to set the destination into the rootfs
- if it's btrfs use go code to copy imports, otherwise use a modified Grab function.

tych0:
A couple of thoughts here:

1. it would be nice if we still bind mounted things in the default configuration (i.e. just `import: foo`); this patch as-is doubles the i/o for an import, copying it once to .stacker and again into the rootfs. Perhaps that means this should be a separate key, like overlay_dirs: in the original issue, so the UX isn't too confusing?

2. It would be really nice if these would just get generated as an OCI layer straight into the output, so that if they didn't change but the contents of the run: section did, you wouldn't get this additional delta.

Let me know if those don't make sense.

lib/file.go
```diff
@@ -31,17 +31,34 @@ func FileCopy(dest string, source string) error {
...
 
-	d, err := os.Create(dest)
+	fi, err := s.Stat()
 	if err != nil {
 		return err
 	}
-	defer d.Close()
 
+	//if preservePermissions {
+	//	err = d.Chmod(fi.Mode())
```
tych0, line 46: I think we do want to preserve the file's mode, just not the owners and such.

peusebiu:
1. I understand your point. By overlay_dirs you mean to tell stacker what dir should be bind-mounted? By default this would be /stacker, and if you give an overlay_dirs value, then that value will be bind-mounted: /value?
2. I don't understand this one. Can you please explain it? You mean to "repack" only the imports as an oci layer?

1 and 2 are mutually exclusive?

tych0:
> Leave things as is but add a new entry in the yaml, overlay_dirs, so that the user knows where the imports should be located.

Yep, although maybe calling them "additional data" or something vs "imports", since that might be confusing.

> Generate an oci layer which should contain only the imports? Does this work? I need to think about it, I currently don't know how to 'bind' this layer to the stacker.yaml layer.

Yeah, this is a bit more complicated. I imagine it working in the overlay sort 
of like the below. Suppose we have:

```
overlay_dirs:
    - source: /my/host/application
      dest: /opt/application
``` 

It would:

1. compute the hash of that dir via the same way we do for imports in
   Put(); if the hash matches what was already there, we can short circuit the 
   rest of this.
2. Generate an OCI layer for this and insert it into the output; unfortunately 
   I think this would still require a copy, because e.g. there's no way to tell 
   `mksquashfs` to "ignore these permissions" (and we don't have an umoci API 
   for it either, though this would be easier to add).
       
   This step would use your existing `copyImportsInRootfs` function
   to instead copy the imports to some tempdir in roots-dir somewhere while
   changing all the owners the right way. Then it would generate a layer via 
   code like this (which should probably live in its own function):
   https://github.com/anuvu/stacker/blob/master/overlay/pack.go#L144
   
   Once it sticks this layer in the output image, it will have to rename the 
   layer to the right sha and update overlay_metadata.json for the thing it's
   building, so that the actual build would mount this layer will contain these 
   files.

   We'll also have to add a cache entry for this, since the build can change if 
   this directory changes.

Let me know if that helps. I realize this is all kind of messy :)

tych0:
It also just occurred to me: I think we can just return an error if someone uses this `overlay_dirs:` key in btrfs mode. It will encourage more people to switch to overlay mode and let us get rid of btrfs mode even faster :)

peusebiu:
Thank you.
I understood your points and I played a little with the overlay code.

I have a question:
When I need to generate this layer, stacker doesn't have yet an oci repository, only the one where it stores layer bases.
What if I modify the base layer, and add this new layer there, then the overlay will pick it up as usual: https://github.com/anuvu/stacker/blob/master/overlay/pack.go#L298
And I will run this before setupRootfs, so that the overlay_dirs will be present in the container while executing the run section.

I think something like this: https://github.com/opencontainers/umoci/blob/07fa845e5b068dee64dcbf391b456a564a6fcfa6/cmd/umoci/insert.go
I might say a stupid thing, all the implications are not yet clear to me.

Otherwise how would I add this layer before executing the run section and before Repack? Because I need those files to be present in the build container.

peusebiu:
I see now, thanks a lot. I'm starting to understand the storage part of the stacker.

tych0:
Feel free to send any refactoring patches you like as well, I'm sure there are things that could be improved.

peusebiu:
Hello Tycho,

What I did:
1) copy imports in tmp dir (./roots/$imageName/overlay_dirs)
2) generate layer from this directory into ./oci repo
3) symlinked tmp dir with imports to overlayPath (roots/sha256_fa2cea153cb781fa2607f9fc6bad85b039dc06526e9c71a2bfc61bf4e173c697) (also tried unpackOne but the generated layer is not compressed, it's just a regular file :-??)
4) modified overlay_metadata, added the new layer Descriptor:
```
{
  "Manifests": {
    "tar": {
      "schemaVersion": 2,
      "config": {
        "digest": "sha256:8dd57e171a61368ffcfde38045ddb6ed74a32950c271c1da93eaddfb66a77e78",
        "size": 1074
      },
      "layers": [
        {
          "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
          "digest": "sha256:7a0437f04f83f084b7ed68ad9c4a4947e12fc4e1b006b38129bac89114ec3621",
          "size": 75181999
        },
        {
          "mediaType": "application/vnd.oci.image.layer.v1.tar",
          "digest": "sha256:fa2cea153cb781fa2607f9fc6bad85b039dc06526e9c71a2bfc61bf4e173c697",
          "size": 59335
        }
      ]
    }
  },
  "BuiltLayers": null,
  "OverlayDirs": [
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar",
      "digest": "sha256:fa2cea153cb781fa2607f9fc6bad85b039dc06526e9c71a2bfc61bf4e173c697",
      "size": 59335
    }
  ]
}

```
Now when the lxc container tries to setup the rootfs I get:
```
lxc rootfs overlay arg overlayfs:/home/peusebiu/unpriv/roots/sha256_fa2cea153cb781fa2607f9fc6bad85b039dc06526e9c71a2bfc61bf4e173c697/overlay:/home/peusebiu/unpriv/roots/sha256_fa2cea153cb781fa2607f9fc6bad85b039dc06526e9c71a2bfc61bf4e173c697/overlay:/home/peusebiu/unpriv/roots/sha256_7a0437f04f83f084b7ed68ad9c4a4947e12fc4e1b006b38129bac89114ec3621/overlay:/home/peusebiu/unpriv/roots/parent/overlay
failed to start container parent
lxc parent 20210409152127.159 ERROR    overlay - storage/overlay.c:ovl_mount:492 - Invalid argument - Failed to mount "/home/peusebiu/unpriv/roots/sha256_fa2cea153cb781fa2607f9fc6bad85b039dc06526e9c71a2bfc61bf4e173c697/overlay:/home/peusebiu/unpriv/roots/sha256_fa2cea153cb781fa2607f9fc6bad85b039dc06526e9c71a2bfc61bf4e173c697/overlay:/home/peusebiu/unpriv/roots/sha256_7a0437f04f83f084b7ed68ad9c4a4947e12fc4e1b006b38129bac89114ec3621/overlay" on "/usr/local/lib/lxc/rootfs" with options "upperdir=/home/peusebiu/unpriv/roots/parent/overlay,lowerdir=/home/peusebiu/unpriv/roots/sha256_fa2cea153cb781fa2607f9fc6bad85b039dc06526e9c71a2bfc61bf4e173c697/overlay:/home/peusebiu/unpriv/roots/sha256_fa2cea153cb781fa2607f9fc6bad85b039dc06526e9c71a2bfc61bf4e173c697/overlay:/home/peusebiu/unpriv/roots/sha256_7a0437f04f83f084b7ed68ad9c4a4947e12fc4e1b006b38129bac89114ec3621/overlay"
lxc parent 20210409152127.159 ERROR    conf - conf.c:lxc_mount_rootfs:1259 - Failed to mount rootfs "overlayfs:/home/peusebiu/unpriv/roots/sha256_fa2cea153cb781fa2607f9fc6bad85b039dc06526e9c71a2bfc61bf4e173c697/overlay:/home/peusebiu/unpriv/roots/sha256_fa2cea153cb781fa2607f9fc6bad85b039dc06526e9c71a2bfc61bf4e173c697/overlay:/home/peusebiu/unpriv/roots/sha256_7a0437f04f83f084b7ed68ad9c4a4947e12fc4e1b006b38129bac89114ec3621/overlay:/home/peusebiu/unpriv/roots/parent/overlay" onto "/usr/local/lib/lxc/rootfs" with options "(null)"
lxc parent 20210409152127.159 ERROR    conf - conf.c:lxc_setup_rootfs_prepare_root:3142 - Failed to setup rootfs for
lxc parent 20210409152127.159 ERROR    conf - conf.c:lxc_setup:3278 - Failed to setup rootfs
lxc parent 20210409152127.159 ERROR    start - start.c:do_start:1218 - Failed to setup container "parent"
lxc parent 20210409152127.160 ERROR    sync - sync.c:__sync_wait:36 - An error occurred in another process (expected sequence number 5)
lxc parent 20210409152127.160 ERROR    start - start.c:__lxc_start:1999 - Failed to spawn container "parent"
error: run commands failed: execute failed: exit status 1
github.com/anuvu/stacker.(*Builder).Build
	/home/peusebiu/stacker/build.go:468
github.com/anuvu/stacker.(*Builder).BuildMultiple
	/home/peusebiu/stacker/build.go:575
main.doBuild
	/home/peusebiu/stacker/cmd/build.go:103
github.com/urfave/cli.HandleAction
	/home/peusebiu/go/pkg/mod/github.com/urfave/cli@v1.22.5/app.go:524
github.com/urfave/cli.Command.Run
	/home/peusebiu/go/pkg/mod/github.com/urfave/cli@v1.22.5/command.go:173
github.com/urfave/cli.(*App).Run
	/home/peusebiu/go/pkg/mod/github.com/urfave/cli@v1.22.5/app.go:277
main.main
	/home/peusebiu/stacker/cmd/main.go:263
runtime.main
	/usr/lib/go-1.15/src/runtime/proc.go:204
runtime.goexit
	/usr/lib/go-1.15/src/runtime/asm_amd64.s:1374
error: exit status 1

github.com/anuvu/stacker/container.MaybeRunInUserns
	/home/peusebiu/stacker/container/container.go:120
main.main.func2
	/home/peusebiu/stacker/cmd/main.go:258
github.com/urfave/cli.(*App).Run
	/home/peusebiu/go/pkg/mod/github.com/urfave/cli@v1.22.5/app.go:264
main.main
	/home/peusebiu/stacker/cmd/main.go:263
runtime.main
	/usr/lib/go-1.15/src/runtime/proc.go:204
runtime.goexit
	/usr/lib/go-1.15/src/runtime/asm_amd64.s:1374
```

mikemccracken:
@peusebiu @tych0 it looks like this was closed without merging, what happened? did we decide not to fix this?

tych0:
I think this is just an old PR, it was merged as overlay_dirs

PR #154, opened 2021-04-12
build: always annotate with the input stacker.yaml

tych0:
I have no idea why we did if/else here, but it seems fairly reasonable to
want both the stacker.yaml and the git hash (even if the git hash could
tell you the stacker yaml, it won't necessarily tell you the value of
--substitutions or whatever that people ran with). Let's include both.


raharper (approved):
\o/
Now with more stacker annotations!

rchamarthy (approved):

PR #155, opened 2021-04-13
Add an overlay_dirs yaml entry

peusebiu:
```
overlay_dirs:
    - source: example_dir
      dest: /path/to/target
```

This will copy source's contents into dest
in the target image, preserving permission,
links, times but not ownership.
source's contents will be packed into a
layer and added to the final image.

### DIFF
--- a/build.go
+++ b/build.go
@@ -407,6 +407,16 @@ func (b *Builder) Build(s types.Storage, file string) error {
 			return err
 		}
 
+		overlayDirs, err := l.ParseOverlayDirs()
+		if err != nil {
+			return err
+		}
+
+		err = s.SetOverlayDirs(name, overlayDirs, opts.LayerTypes)
+		if err != nil {
+			return err
+		}
+
 		c, err := NewContainer(opts.Config, s, name)
 		if err != nil {
 			return err
--- a/lib/dir.go
+++ b/lib/dir.go
@@ -1,0 +1,62 @@
+package lib
+
+import (
+	"io/ioutil"
+	"os"
+	"path"
+
+	"github.com/pkg/errors"
+)
+
+// DirCopy copies a whole directory recursively
+func DirCopy(dest string, source string) error {
+
+	var err error
+	var fds []os.FileInfo
+	var srcinfo os.FileInfo
+
+	if srcinfo, err = os.Stat(source); err != nil {
+		return errors.Wrapf(err, "Coudn't stat %s", source)
+	}
+
+	linkFI, err := os.Lstat(source)
+	if err != nil {
+		return errors.Wrapf(err, "Coudn't stat link %s", source)
+	}
+
+	// in case the dir is a symlink
+	if linkFI.Mode()&os.ModeSymlink != 0 {
+		target, err := os.Readlink(source)
+		if err != nil {
+			return errors.Wrapf(err, "Coudn't read link %s", source)
+		}
+
+		return os.Symlink(target, dest)
+	}
+
+	dirMode := srcinfo.Mode()
+
+	if err = os.MkdirAll(dest, dirMode); err != nil {
+		return errors.Wrapf(err, "Coudn't mkdir %s", dest)
+	}
+
+	if fds, err = ioutil.ReadDir(source); err != nil {
+		return errors.Wrapf(err, "Coudn't read dir %s", source)
+	}
+
+	for _, fd := range fds {
+		srcfp := path.Join(source, fd.Name())
+		dstfp := path.Join(dest, fd.Name())
+
+		if fd.IsDir() {
+			if err = DirCopy(dstfp, srcfp); err != nil {
+				return err
+			}
+		} else {
+			if err = FileCopy(dstfp, srcfp); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
--- a/test/overlay-dirs.bats
+++ b/test/overlay-dirs.bats
@@ -1,0 +1,122 @@
+load helpers
+
+function setup() {
+    stacker_setup
+}
+
+function teardown() {
+    cleanup
+    rm -rf recursive bing.ico || true
+}
+
+@test "overlay_dirs works" {
+    require_storage overlay
+    mkdir dir_to_overlay
+    touch dir_to_overlay/file1
+    touch dir_to_overlay/file2
+    touch dir_to_overlay/executable
+    chmod +x dir_to_overlay/executable
+    cat > stacker.yaml << EOF
+first:
+    from:
+        type: oci
+        url: $CENTOS_OCI
+    overlay_dirs:
+        - source: dir_to_overlay
+    run: |
+        [ -f /file1 ]
+        [ -f /file2 ]
+        [ -x /executable ]
+EOF
+    stacker build
+}
+
+@test "import from overlay_dir works" {
+    require_storage overlay
+    mkdir dir_to_overlay
+    touch dir_to_overlay/file
+    cat > stacker.yaml << EOF
+first:
+    from:
+        type: oci
+        url: $CENTOS_OCI
+    overlay_dirs:
+        - source: dir_to_overlay
+    run: |
+        [ -f /file ]
+second:
+    from:
+        type: built
+        tag: first
+    import: stacker://first/file
+    run: |
+        [ -f /stacker/file ]
+EOF
+    stacker build
+}
+
+@test "overlay_dirs dest works" {
+    require_storage overlay
+    mkdir dir_to_overlay
+    touch dir_to_overlay/file
+    cat > stacker.yaml << EOF
+first:
+    from:
+        type: oci
+        url: $CENTOS_OCI
+    overlay_dirs:
+        - source: dir_to_overlay
+          dest: /usr/local
+    run: |
+        [ -f /usr/local/file ]
+EOF
+    stacker build
+}
+
+@test "overlay_dirs cache works" {
+    require_storage overlay
+    mkdir dir_to_overlay
+    touch dir_to_overlay/file
+    cat > stacker.yaml << EOF
+first:
+    from:
+        type: oci
+        url: $CENTOS_OCI
+    overlay_dirs:
+        - source: dir_to_overlay
+          dest: /usr/local
+    run: |
+        [ -f /usr/local/file ]
+EOF
+    stacker build
+    stacker build
+    echo $output | grep "found cached layer first"
+    echo "modifying file" > dir_to_overlay/file
+    stacker build
+    echo $output | grep "cache miss because overlay_dir content changed"
+    rm -rf roots
+    stacker build
+    echo $output | grep "cache miss because overlay_dir was missing"
+}
+
+@test "overlay_dirs don't preserve ownership" {
+    require_storage overlay
+    mkdir dir_to_overlay
+    touch dir_to_overlay/file
+    touch dir_to_overlay/file2
+    chown 1234:1234 dir_to_overlay/file
+    cat > stacker.yaml << EOF
+first:
+    from:
+        type: oci
+        url: $CENTOS_OCI
+    overlay_dirs:
+        - source: dir_to_overlay
+          dest: /usr/local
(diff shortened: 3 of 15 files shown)


peusebiu:
Hello Tycho.

Tried your approach from the last PR (to generate the layer before executing the run section). The issue with this approach, for me, is to add this layer into the final image. Because when I first generate it, I get a Descriptor{} from umoci API, and if I try to add it in the final image, Descriptor values changes, mainly the Digest, and at this point I already have the layer unpacked into overlay/sha256_$DIGEST, $DIGEST being different now that I've added the layer into the final image.

Hope this is clear enough...

So instead of your approach I did:

1) copy imports into a temporary directory
2) mount this directory into lxc container (via GetLXCRootfsString)
3) Execute the run section
4) Generate a layer (one per overlay_dir) from the temporary directory
5) Symlink temporary directory contents to overlayPath() of 4) layer
5) Run the rest of the repack

Is this approach good? I want to know if I can proceed with writing tests. Also I didn't create a History for these  additional layers, I also need to do that.

I've observed that your code takes care to create another layer if I delete some files from overlay_dirs.

lib/dir.go
```diff
@@ -0,0 +1,60 @@
...
+
+// DirCopy copies a whole directory recursively
+func DirCopy(dest string, source string) error {
+
+	var err error
+	var fds []os.FileInfo
+	var srcinfo os.FileInfo
+
+	if srcinfo, err = os.Stat(source); err != nil {
+		return err
```
tych0, line 17: Looks like there's a bunch of filesystem operations here and in a few other functions which don't use errors.Wrapf(), which prevents us from getting stack traces from the errors making things harder to debug. Can you add a Wrapf() from these filesystem operations?

You don't need to add Wrapf() to errors which are already wrapped (aka ones that come from stacker or umoci APIs), just ones that won't include stack trace info.

import.go
```diff
@@ -355,3 +355,26 @@ func Import(c types.StackerConfig, storage types.Storage, name string, imports t
 
 	return nil
 }
+
+// Copy imports to a container rootfs
+func copyImportsInRootfs(name string, overlayDirs types.OverlayDirs, sc types.StackerConfig) error {
```
tych0, line 360: I don't think this function name really makes sense, since it's not actually copying things into the rootfs. Maybe "copyOverlayDir" or similar?

build.go
```diff
@@ -405,6 +405,26 @@ func (b *Builder) Build(s types.Storage, file string) error {
...
 
+		if len(l.OverlayDirs) > 0 {
+			// get absolute paths for overlay_dirs source
+			overlayDirs, err := l.ParseOverlayDirs()
+			if err != nil {
+				return err
+			}
+
+			// copy overlay_dirs contents into a temporary dir in roots
+			err = copyImportsInRootfs(name, overlayDirs, opts.Config)
```
tych0, line 416: I wonder if this and the related code can't all be moved into `overlay/`, since that's the only place it is used?

test/overlay-dirs.bats
```diff
@@ -0,0 +1,98 @@
...
+    cleanup
+    rm -rf recursive bing.ico || true
+}
+
+@test "overlay_dirs works" {
+    require_storage overlay
+    mkdir dir_to_overlay
+    touch dir_to_overlay/file1
+    touch dir_to_overlay/file2
+    touch dir_to_overlay/executable
```
tych0, line 17: Can you add a test here or somewhere that explicitly tests that we do *not* copy owner + group?

tych0:
> The issue with this approach, for me, is to add this layer into the final image. Because when I first generate it, I get a Descriptor{} from umoci API, and if I try to add it in the final image, Descriptor values changes, mainly the Digest, and at this point I already have the layer unpacked into overlay/sha256_$DIGEST, $DIGEST being different now that I've added the layer into the final image.

I'm not sure I understand what is going on here. The descriptor's shouldn't change once you've generated the layer, both because overlay should prevent you from writing additional data to this directory and that the descriptor itself shouldn't be mutable.

The way you currently have it organized seems like it should "Just Work" to do it before the run section, unless I'm missing something? Can you describe where in the code the descriptor is actually changed?

import.go
peusebiu: Ok, I will change

build.go
peusebiu: You are right, I will move it there, thanks.

test/overlay-dirs.bats
peusebiu: Good idea, I thought about that, but forgot in the meantime.

peusebiu:
Let me show you the code from the last week.

What I meant is:
1) create the layer before run: https://github.com/anuvu/stacker/compare/master...peusebiu:overlay_dirs_gen_b4_run?expand=1#diff-65d308d7a976a5e84864580ddc3771d0108ed7c1b92314da3cbe02a1ae89c3e2R209
2) after run, at repack, I tried to add this layer to the final image: https://github.com/anuvu/stacker/compare/master...peusebiu:overlay_dirs_gen_b4_run?expand=1#diff-65d308d7a976a5e84864580ddc3771d0108ed7c1b92314da3cbe02a1ae89c3e2R556

But as you can see I get a different descriptor object from the mutator.
I'm sure I am missing something, and there must be a way to add the layer without generating a new Descriptor.

lib/dir.go
peusebiu: Ok, got it. Thanks

tych0:
> I'm sure I am missing something, and there must be a way to add the layer without generating a new Descriptor.

Ah, yes, now I see. The umoci API for this is not very good (it only wants you to add things via it, but we do it externally). One approach I've used in the past is to edit the manifest "by hand": https://github.com/anuvu/stacker/blob/master/oci/oci.go#L60

I guess you can do something like this by-hand editing in initializeBasesInOutput()?

peusebiu:
> > I'm sure I am missing something, and there must be a way to add the layer without generating a new Descriptor.
> 
> Ah, yes, now I see. The umoci API for this is not very good (it only wants you to add things via it, but we do it externally). One approach I've used in the past is to edit the manifest "by hand": https://github.com/anuvu/stacker/blob/master/oci/oci.go#L60
> 
> I guess you can do something like this by-hand editing in initializeBasesInOutput()?

I don't understand, why in initializeBasesInOutput()?  Maybe that's what I'm missing now.

I tried to do it in the repack but if I modify the image before init-ing the mutators, then the descriptorPaths will be of length 0 : 
https://github.com/anuvu/stacker/pull/155/files#diff-65d308d7a976a5e84864580ddc3771d0108ed7c1b92314da3cbe02a1ae89c3e2R481

If I do it after mutators then they will overwrite my oci repo to not include the overlay_dir layer.

tych0:
> The question is what am I doing wrong that len(descriptorPaths) == 0 ?

This suggests to me that the output isn't getting correctly written somehow, even if you're not adding the overlay_dirs correctly, it should still return at least the original oci base image (or an empty image in case of a tar source, for example). Let me play with the code a bit.

tych0:
Ok, I finally got around to looking at this, looks like there are three distinct bugs here. AddBlobByDescriptor() uses the wrong media type in two places:

```
diff --git a/oci/oci.go b/oci/oci.go
index 7777b27..d1941c7 100644
--- a/oci/oci.go
+++ b/oci/oci.go
@@ -94,7 +94,7 @@ func AddBlobByDescriptor(oci casext.Engine, name string, desc ispec.Descriptor)
        }
 
        manifest.Config = ispec.Descriptor{
-               MediaType: desc.MediaType,
+               MediaType: ispec.MediaTypeImageConfig,
                Digest:    configDigest,
                Size:      configSize,
        }
@@ -105,7 +105,7 @@ func AddBlobByDescriptor(oci casext.Engine, name string, desc ispec.Descriptor)
        }
 
        desc = ispec.Descriptor{
-               MediaType: ispec.MediaTypeImageConfig,
+               MediaType: ispec.MediaTypeImageManifest,
                Digest:    manifestDigest,
                Size:      manifestSize,
        }
```

And the destination path for overlay_dirs was computed incorrectly when generating the layers:

```
diff --git a/overlay/overlay-dirs.go b/overlay/overlay-dirs.go
index 71a5924..920034b 100755
--- a/overlay/overlay-dirs.go
+++ b/overlay/overlay-dirs.go
@@ -50,7 +50,7 @@ func generateOverlayDirLayer(name string, layerType types.LayerType, overlayDir
        }
        defer oci.Close()
 
-       contents := path.Join(config.RootFSDir, name, "overlay_dirs", path.Base(overlayDir.Source), overlayDir.Dest)
+       contents := path.Join(config.RootFSDir, name, "overlay_dirs", path.Base(overlayDir.Source))
 
        var blob io.ReadCloser
        if layerType == "tar" {
```

With these two fixed, the overlay-dirs tests pass for me (though it has dos line endings, would be good to convert those to unix).

It also looks like there's some code duplication in the layer generation/adding code in overlay/overlay-dirs.go with the existing layer adding code in overlay/pack.go that could be abstracted.

peusebiu:
Thanks for taking your time to test my changes, and for the in depth review. 
Can you take a look again? I'm not 100% sure about its "design" so if you have any suggestion I will update accordingly.

I need to also add docs and some comments.  I will do it soon.

types/layer.go
```diff
@@ -73,10 +80,10 @@ func getImportFromInterface(v interface{}) (Import, error) {
 	m2, ok := v.(map[string]interface{})
 	if ok {
 		// check for nil hash so that we won't end up with "nil" string values
-		if m["hash"] == nil {
+		if m2["hash"] == nil {
 			hash = ""
 		} else {
-			hash = fmt.Sprintf("%v", m["hash"])
+			hash = fmt.Sprintf("%v", m2["hash"])
```
tych0, line 86: This seems like a bug fix, can you add it as a separate commit with its own message?

tych0:
Over all this looks pretty good. Looks like you need some quoting in the tests to make it work, and it looks like there's a chmod +x of go.mod that doesn't make sense, but other than that the fixing the review comments above works for me.

peusebiu:
Something strange is happening here:
# + '[' -f /usr/local/file ']'
# + owner=root
# + group=root
# + '[' '' = root ']'
# error: run commands failed: execute failed: exit status 1

I tried all types of strings checking:
 [[ "${owner}" == "root" ]] 
[ "$owner" = "root" ]
[ "$owner" == "root" ]

I might miss something...

tych0:
Yeah, you're hitting kind of an awful weirdness between bash heredoc quoting and our stacker.yaml literals. I wish there was a better way to do this, but I can get it to pass with:

```
diff --git a/test/overlay-dirs.bats b/test/overlay-dirs.bats
index ad5db9b..1f32bae 100755
--- a/test/overlay-dirs.bats
+++ b/test/overlay-dirs.bats
@@ -115,10 +115,8 @@ first:
           dest: /usr/local
     run: |
         [ -f /usr/local/file ]
-        owner=$(ls -ld /usr/ | awk '{print $3}')
-        group=$(ls -ld /usr/ | awk '{print $4}')
-        [ "$owner" = "root" ]
-        [ "$group" = "root" ]
+        [ "$(stat --format=%U /usr)" == "root" ]
+        [ "$(stat --format=%G /usr)" == "root" ]
 EOF
     stacker build
 }
```

tych0:
Stuff in test brackets `[]` is evaluated at execution time, but stuff *not* there is evaluated at heredoc writing time :(

peusebiu:
Understood, thanks.
Your example work on my machine only if I escape the dollar in the if expr
[ "\$(stat --format=%G /usr)" == "root" ]

tych0:
Oh, so maybe I'm wrong and they're *all* evaluated at heredoc time. :(

lib/dir.go
tych0: Seems like we still need a bunch of Wrapf()s here?

tych0: Or at least WithStack()

tych0:
Awesome, thanks for all your work on this!

PR #157, opened 2021-04-20
Inform user that cached layers are rebuilt if bind mounts are used

raharper:
Stacker will rebuild a cached layer if it uses bind mounts.  Emit an
info message so users understand stacker's behavior.

- Add a check in the cache test for the logged message

### DIFF
--- a/build.go
+++ b/build.go
@@ -398,6 +398,8 @@ func (b *Builder) Build(s types.Storage, file string) error {
 
 				log.Infof("missing some cached layer output types, building anyway")
 			}
+		} else if len(binds) > 0 {
+			log.Infof("rebuilding cached layer due to use of binds in stacker file")
 		}
 
 		err = SetupRootfs(baseOpts)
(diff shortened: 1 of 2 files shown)


tych0:
There was a small logic error, and then also the regex wasn't matching, but this grep seems to work. I have no idea why the regex wasn't working and the one below it was, but grep works just the same...

tych0:
Anyway, patch lgtm assuming CI likes it.

build.go
```diff
@@ -398,6 +398,8 @@ func (b *Builder) Build(s types.Storage, file string) error {
 
 				log.Infof("missing some cached layer output types, building anyway")
 			}
+		} else if len(binds) > 0 {
```
raharper, line 401: Hrm, I had the cached && binds as I didn't want to emit the message if we didn't have a cache hit ... the first time building a layer that has binds we'll emit this message now, right?

PR #158, opened 2021-04-21
build: only emit rebuild message if cached and has binds

raharper:
We only want to emit the rebuildig message if the layer has been
built once and we have a cache hit.  Also switch back to bash pattern
matching in the test-case.  The previous attempt didn't account for the
fact that the output is a single line so matching a substring requires
accepting anything before and after the match.


raharper:
Let's see if I've fixed the pattern matching in the testcase ... if not I'll re-push the grep

PR #160, opened 2021-04-21
dependency calculation: print missing deps

tych0:
It is annoying to have stacker say "I can't figure out what your deps are"
without telling you what it couldn't resolve. Let's have it print these.


raharper (approved):
Thanks! This is a really nice user-friendly improvement.

PR #162, opened 2021-05-24
lib: add copy option to set manifest type

### LINKED ISSUE #161: add option to set manifest type while copying images
If we want to copy image from docker registry to oci registry, there should be an option to force manifest type so that we can make sure oci compatible blobs are transferred to oci registry. 

Similar to ```skopeo copy --format=oci```

shimish2:
Closes #161

### DIFF
--- a/lib/image.go
+++ b/lib/image.go
@@ -46,16 +46,17 @@ func localRefParser(ref string) (types.ImageReference, error) {
 }
 
 type ImageCopyOpts struct {
-	Src          string
-	SrcUsername  string
-	SrcPassword  string
-	Dest         string
-	DestUsername string
-	DestPassword string
-	SrcSkipTLS   bool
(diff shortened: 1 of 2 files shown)


tych0:
Can you add a link showing what valid values of this string are; and perhaps a test?

lib/image.go
```diff
@@ -113,6 +114,13 @@ func ImageCopy(opts ImageCopyOpts) error {
 		}
 	}
 
+	// Set ForceManifestMIMEType
+	// Supported manifest type :- https://github.com/anuvu/image/blob/master/manifest/manifest.go#L35
```
tych0, line 118: Can you link to the upstream containers/image here?

tych0: Ping?

shimish2: Updated link, please review

PR #174, opened 2021-06-23
tar base layers: fix computation of dependency order

tych0:
The problem here is that we explicitly allow and document stacker:// style
imports for the url: portion of tar: base layers, so people can compose
tarballs in one layer and emit them in other layers.

However, the dependency order calculation did not take into account this
dependency. Let's add it to the computation.

Additionally, we fix a comment on DependencyOrder(), since it *does* in
fact re-order things now (and the comment had some poor grammar...)

### DIFF
--- a/test/dependency-order.bats
+++ b/test/dependency-order.bats
@@ -34,3 +34,59 @@ EOF
     bad_stacker build
     echo "${output}" | grep "couldn't find dependencies for test: stacker://foo/bar, stacker://baz/foo"
 }
+
+@test "stacker:// style nesting w/ type built works" {
+    cat > stacker.yaml <<EOF
+minbase1:
+  from:
+    type: oci
+    url: $UBUNTU_OCI
+  run: |
+    echo pkgtool install cpio pigz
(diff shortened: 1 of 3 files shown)


test/dependency-order.bats
```diff
@@ -34,3 +34,66 @@ EOF
...
+    build_only: true
+    from:
+        type: built
+        tag: minbase1
+    run: |
+        #!/bin/bash -ex
+        mkdir /output
+        touch /output/installer-initrd-base.cpio
+
+        echo "${STACKER_LAYER_NAME}" > /layer.txt
```
smoser, line 57: you probably need to escape this if you want it to work.
I would expect that the STACKER_LAYER_NAME variable is rendered at time it is read.
It looks like you *want* that to happen as `$UBUNTU_OCI` probably needs to be rendered from the current environment.

tych0: I think we can just drop this actually, it's copypasta from your minified example that's not critical to the test.

PR #192, opened 2021-08-17
ci: run with go 1.17 too

tych0:
mostly it'll be interesting to see if the new register allocator will be
substantially faster for all of our pure go tar/gzip usage, which is mostly
what dominates stacker's cpu time.


tych0:
I guess not, probably because things are i/o bound. Need a more scientific test.

Issue #206, opened 2021-10-20
allow pinning shasum of imports


mikemccracken:
Motivation: building with a given stackerfile should always produce the same results.

```yaml
import:
  - source: http://$artifactory/place/foo.bin
    dest: /bin/foo
    sha256: 300abcdef...
```

if `sha256` exists, then we check it and fail the build if the source doesn't match it.

we can also add a flag `--remote-input-hash-required` (bad name) that fails the build if any remote imports do not have the `sha256` key. because local imports are likely to be built from the same commit that contains the stacker file, we do not want to force a commit between binary builds and image builds.

we could try to be even smarter about what counts as a local import, where stacker would only count it as local if it's in the same git tree as the stacker file

we could also incorporate #41 along with this since we need to change the syntax anyway

tsnaik:
> we can also add a flag --remote-input-hash-required (bad name) that fails the build if any remote imports do not have the sha256 key.

Does something like this seem good?

```
--require-hash                 require each remote import to have a hash provided in stackerfiles
```

tych0:
If I'm honest I don't know that I love the optionality for this. If people have specified a hash and it's different, then the build should fail. If the hash is expected to change, they should change the hash.

I think that's what your patch in #211 did, so I'm going to go ahead and close this.

mikemccracken:
@tych0 the extra flag was not to avoid checking hashes, but to tell stacker to fail the build during stackerfile parsing if it finds an import that does not have a hash specified in the file.

we want to enforce a rule that all stackerfiles in a repo must specify hashes.

tych0:
I see, makes sense, thanks for the clarification.

PR #211, opened 2021-10-29
import: add a hash check for http(s) imports 

tsnaik:
This builds on top of 60cb730
Before this commit it only checked  HTTP header "X-Checksum-Sha256" with the hash provided in stacker.yaml

Now it also verifies the sha256 of the newly downloaded file on top of existing checks

Also resolves ~part1 of #206

### DIFF
--- a/network.go
+++ b/network.go
@@ -16,7 +16,7 @@ import (
 )
 
 // download with caching support in the specified cache dir.
-func Download(cacheDir string, url string, progress bool, remoteHash string, remoteSize string) (string, error) {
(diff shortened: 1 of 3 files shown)


network.go
```diff
@@ -76,7 +76,6 @@ func Download(cacheDir string, url string, progress bool, remoteHash string, rem
 		os.RemoveAll(name)
 		return "", errors.Errorf("couldn't download %s: %s", url, resp.Status)
 	}
-
```
tych0, line 79: looks like a stray whitespace change here?

tych0:
One small nit, but other than that this looks great, thank you!

network.go
tsnaik: fixed. thanks

PR #213, opened 2021-11-02
Check environment

tych0:


mikemccracken (approved):

rchamarthy (approved):

btrfs/btrfs.go
```diff
@@ -517,3 +517,37 @@ func (b *btrfs) SetOverlayDirs(name string, overlayDirs types.OverlayDirs, layer
...
+	// can we mount btrfs filesystems? this is not allowed as an
+	// unprvileged user in any kernel, so we need to be euid 0 and btrfs
+	// needs to be listed in /proc/filesystems.
+	if os.Geteuid() != 0 {
+		return errors.Errorf("not root and %v is not btrfs", config.RootFSDir)
+	}
+
+	// btrfs might be built as a module, so let's modprobe; modprobe might
+	// fail if it was builtin, so let's just ignore modprobe's result
+	_ = exec.Command("modprobe", "btrfs").Run()
```
smoser, line 541: just to avoid unnecessary failing modprobe, you could do this only if not already FilesystemIsSupported("btrfs").

```go

	if hasBtrfs, err = mount.FilesystemIsSupported("btrfs"); err != nil {
		return err
	} else if !hasBtrfs {
		_ = exec.Command("modprobe", "btrfs").Run()
		if hasBtrfs, err = mount.FilesystemIsSupported("btrfs"); err != nil {
			return err
		}
	}
```

smoser (commented):
overall very nicely done.
Just one suggestion inline to avoid unnecessary modprobe.  And probably this allows you to not ignore the return value of Run(), as at the poitn where you got there.... that was the only way to fix.

btrfs/btrfs.go
tych0: yep, that's a nice improvement, thanks.

smoser (approved):
i like it.

PR #214, opened 2021-11-03
import: add flag to verify hash in remote imports

tsnaik:
when stacker build is called with the flag `--require-hash`, it will fail the build if any remote import (http/https) doesn't have hash in its yaml entry.

```
--require-hash                  require all remote imports to have a hash provided in stackerfiles
```

*It still allows local or stacker imports without the hash*

Closes anuvu/stacker#206

### DIFF
--- a/build.go
+++ b/build.go
@@ -28,6 +28,7 @@ type BuildArgs struct {
 	OnRunFailure string
 	LayerTypes   []types.LayerType
 	OrderOnly    bool
+	HashRequired bool
 	SetupOnly    bool
 	Progress     bool
 }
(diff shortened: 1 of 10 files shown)


import.go
```diff
@@ -227,23 +227,28 @@ func acquireUrl(c types.StackerConfig, storage types.Storage, i string, cache st
 
 	// It's just a path, let's copy it to .stacker.
 	if url.Scheme == "" {
-		return importFile(i, cache, expectedHash)
+		return importFile(importPath, cache, expectedHash)
 	} else if url.Scheme == "http" || url.Scheme == "https" {
 		// otherwise, we need to download it
+
+		if hashRequired && expectedHash == "" {
+			return "", errors.Errorf("Remote import needs a hash in yaml for path: %s", importPath)
```
tych0, line 235: can we do this validation in NewStackerFile() so that it fails faster vs. halfway through the build?

tych0:
one small rearrangement, but overall looks good, thanks!

import.go
tsnaik: Done. Thanks!

tsnaik:
> one small rearrangement, but overall looks good, thanks!

Made the requested changes. Thanks.

tsnaik:
(pushed doc changes for this)

tych0:
Great, thanks!

PR #225, opened 2021-12-01
Hash noncollision

tych0:


tych0:
Seems as though this doesn't *quite* fix things. I think the right fix is really to resolve these paths completely in NewStackerfile() so we don't need to keep track of referenceDirectory at all. I will work on that at some point, but we have more pressing issues elsewhere...

base.go
```diff
@@ -89,7 +89,7 @@ func SetupRootfs(o BaseLayerOpts) error {
 	}
 }
 
-func importContainersImage(is *types.ImageSource, config types.StackerConfig, progress bool) error {
+func importContainersImage(is types.ImageSource, config types.StackerConfig, progress bool) error {
```
hallyn, line 92: it's probably safer this way then, but does this have a non-negligable performance impact due to full struct copy?

tych0: It shouldn't, and like I say in the commit message, there is some go creedo somewhere that I read that says "don't use pointers unless you intend to mutate data". In fact it might be faster due to less dereferences...

hallyn: One dereference at function call cannot be slower than allocating and copying the struct to a new location :)

But I'll buy that it's the "proper" thing to do, so resolving this, thanks.

hallyn (approved):
LGTM -  I don't know based on your comment whether you want to merge this in the meantime, or wait until you have more time to look back into this, so I'll just approve but not merge.

base.go
tych0: > One dereference at function call cannot be slower than allocating and copying the struct to a new location :)

I don't agree, but I also don't think it matters.

tych0:
I don't want to merge this as-is, since it doesn't fix the bug. Three of the commits are fine I think, but the one about ReferenceDirectory is unnecessary and doesn't help anything. Really we should just be dropping ReferenceDirectory entirely, which I'm working on now.

tych0:
Ok, I dropped the ReferenceDirectory commit. Let me know what you think of these.

hallyn (approved):

PR #226, opened 2021-12-02
Fix make check

tristanm-cisco:
Make check with a newer golangci-lint flagged some valid issues.

Channels for signal propagation need to be buffered.  Otherwise a dead lock can occur as unbuffered channels block the sender until the receiver checks for the message.  Real bad behavior if for instance the signal is on the same kernel thread as the go routine that's waiting for the signal.

An err variable was ignored.  The change simply prints the error and continues to maintain the existing semantics but adding debugging bread crumbs.  I welcome alternative suggestions for expected behavior.

Updated the docs with what was needed for the development lifecycle on a fairly fresh Ubuntu 20.04.  The ppa hasn't been updated for focal, but doesn't appear to need to be required.  Several packages commonly already installed weren't listed.  Specifics of how squashtool needs to be installed for `make check` added.

Fun fact:  if btrfs-progs isn't installed, a `make check` crashes the my Ubuntu's graphical env and dumps me out to a login prompt.  Might want to look at why at some point, but didn't see anything obvious in the regular logs.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -67,7 +67,7 @@ func stackerResult(err error) {
 }
 
 func main() {
-	sigquits := make(chan os.Signal)
+	sigquits := make(chan os.Signal, 1)
 	go func() {
 		for range sigquits {
 			debug.PrintStack()
@@ -276,7 +276,7 @@ func main() {
 			cmd = append(cmd[:2], cmd[1:]...)
 			cmd[1] = "--internal-userns"
 
-			forward := make(chan os.Signal)
+			forward := make(chan os.Signal, 3)
 			signal.Notify(forward, syscall.SIGTERM, syscall.SIGHUP, syscall.SIGINT)
 
 			c, err := container.MaybeRunInUserns(cmd)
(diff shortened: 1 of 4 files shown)


doc/install.md
```diff
@@ -31,13 +31,19 @@ https://golang.org/doc/install#install
 The other build dependencies can be satisfied with the following command and
 packages:
 
-    sudo apt install lxc-dev libacl1-dev libgpgme-dev libcap-dev libseccomp-dev libpam0g-dev
+    sudo apt install lxc-dev libacl1-dev libgpgme-dev libcap-dev libseccomp-dev
```
hallyn, line 34: Put a \ at end of this line to make it clear it's continued?

cmd/main.go
```diff
@@ -276,10 +276,13 @@ func main() {
 			cmd = append(cmd[:2], cmd[1:]...)
 			cmd[1] = "--internal-userns"
 
-			forward := make(chan os.Signal)
+			forward := make(chan os.Signal, 3)
 			signal.Notify(forward, syscall.SIGTERM, syscall.SIGHUP, syscall.SIGINT)
 
 			c, err := container.MaybeRunInUserns(cmd)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Error in MaybeRunInUserns(cmd: %s): %s", cmd, err)
```
tych0, line 284: Can we return this error here instead of just printing it so we get a full stack trace?

tych0: I just went ahead and force pushed this fix since I'd also like to drop the hard-coded version as well. Thanks for these!

PR #227, opened 2021-12-02
Switch uses of sha hasher to version which uses hardware accel when available

tristanm-cisco:
Go standard library crypto/sha256 does not use sha2 hardware instructions or AVX-512 when available.  This patch increases byte/cycle hash rate by 4x-8x by using a drop in replacement library.

This adds a new library that is Apache 2.0 licensed, and updates a dependency library that is MIT licensed.

This came out of looking for low hanging hash rate speed ups after running into problems with this issue:  https://github.com/anuvu/stacker/issues/210

### DIFF
--- a/btrfs/utils.go
+++ b/btrfs/utils.go
@@ -1,9 +1,9 @@
 package btrfs
 
 import (
-	"crypto/sha256"
 	"fmt"
 
+	"github.com/minio/sha256-simd"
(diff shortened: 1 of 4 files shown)


rchamarthy (approved):

tych0:
Thanks for this -- I wonder if a similar patch for go-mtree would be useful, since that's what's doing some of our dependency hashing. In any case this is certainly an improvement.

PR #228, opened 2021-12-03
mksquashfs should use zstd when available

tristanm-cisco:
Check for presence of zstd support by looking at mksquashfs --help.  On any failures assume that zstd isn't supported, since mqsqaushfs makes it a little difficult to tell if mksquashfs --help succeeded by exiting with status code 1.

Partially addresses: https://github.com/anuvu/stacker/issues/112 tar support has some open questions, so it'll be in another patch.

Open question:  is there a reason to add a command line flag to force using a certain compression algorithm, whether zstd or gzip?

### DIFF
--- a/squashfs/squashfs.go
+++ b/squashfs/squashfs.go
@@ -11,6 +11,7 @@ import (
 	"os/exec"
 	"path"
 	"strings"
+	"sync"
 
 	stackermtree "github.com/anuvu/stacker/mtree"
 	stackeroci "github.com/anuvu/stacker/oci"
@@ -23,6 +24,9 @@ import (
 	"github.com/vbatts/go-mtree"
 	"golang.org/x/sys/unix"
 )
+
+var checkZstdSupported sync.Once
+var zstdIsSuspported bool
 
 // ExcludePaths represents a list of paths to exclude in a squashfs listing.
 // Users should do something like filepath.Walk() over the whole filesystem,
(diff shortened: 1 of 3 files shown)


squashfs/squashfs.go
```diff
@@ -321,3 +328,24 @@ func whichSearch(name string, paths []string) string {
...
+func mksquashfsSupportsZstd() bool {
+	checkZstdSupported.Do(func() {
+		var stdoutBuffer strings.Builder
+		var stderrBuffer strings.Builder
+
+		cmd := exec.Command("mksquashfs", "--help")
+		cmd.Stdout = &stdoutBuffer
+		cmd.Stderr = &stderrBuffer
+
+		// Ignore errs here as `mksquashfs --help` exit status code is 1
```
darshilgit, line 341: interesting, is this something common?

tristanm-cisco: More common than I'd like.  It's an easy trap to fall into when you're writing a utility like this and want to combine the '--help' flow with the more common in your code dispatches to 'malformed inputs so print usage and error exit'.  mkquashfs is not the only one doing this by any means.

tristanm-cisco:
The previous failure stemmed from zstd support not being present by default in ubuntu 20.04, leading to squashtool ./configureing out the support.  This gives me hesitancy on the design at the moment, given the apparent lack of ubiquity of zstd support.

tych0:
> Open question: is there a reason to add a command line flag to force using a certain compression algorithm, whether zstd or gzip?

I think so, just in case someone knows they're targeting older tooling.

> The previous failure stemmed from zstd support not being present by default in ubuntu 20.04, leading to squashtool ./configureing out the support. This gives me hesitancy on the design at the moment, given the apparent lack of ubiquity of zstd support.

Looks like installing the library fixed it though? That doesn't necessarily scare me that much, although I do think perhaps we should leave the default at gzip for a while.

tych0:
I wonder if we can add a patch to squashtool to list its supported compression as well, so we can make sure both mksquashfs and squashtool support zstd or otherwise fail?

tych0:
I'm gonna close this; I've cherry picked these patches to a local branch I have re-working some of the squashfs stuff, and my plan for compression is to just append it OCI tar style with "+zstd" or "+gzip" to the defined squashfs media type.

PR #242, opened 2022-02-02
deps: revert to our containers/image fork

tych0:
This is effectively a revert of 81bb45693a67 ("deps: move back to upstream
containers/image"), but with a newer containers/image.

It turns out our tests do not test uploading (i.e. transitioning) squashfs
images across OCI -> docker, which breaks with this change.


smoser (approved):
sounds good to me.

PR #257, opened 2022-03-16
ci: add go 1.18 to CI

tych0:


tych0:
seems like there are some problems for golangci-lint in go 1.18 (not to mention github actions was broken when this ran yesterday :); I'll close this until those are resolved.

PR #258, opened 2022-03-17
build: also build libdevmapper in parallel

tych0:
libdevmapper isn't huge, but every little bit helps...

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -36,7 +36,7 @@ build-env:
     tar -xf /stacker/v2_03_15.tar.gz
     cd lvm2-*
     ./configure --enable-static_link
-    make install_device-mapper
+    make -j$(grep -c processor /proc/cpuinfo) install_device-mapper
     cd /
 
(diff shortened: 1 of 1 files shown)


hallyn:
How the heck did this fail?

tych0:
I... don't know :). Seems like maybe they don't really support parallel builds? But I was just looking at the output of some strcpy buffers overflowing. Seems like it might be worth sending them a few patches.

rchincha:
/usr/lib/gcc/x86_64-alpine-linux-musl/11.2.1/../../../../x86_64-alpine-linux-musl/bin/ld: /usr/lib/gcc/x86_64-alpine-linux-musl/11.2.1/../../../../lib/crt1.o: in function `_start_c':
/home/buildozer/aports/main/musl/src/v1.2.2/crt/crt1.c:18: undefined reference to `main'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:66: dmsetup.static] Error 1

PR #272, opened 2022-04-01
Automatically create releases on new tags

peusebiu:
example automatic release: https://github.com/peusebiu/stacker/releases/tag/v1.0.3


.github/workflows/release.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up golang
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17.x
+      - name: install dependencies
+        run: |
+          sudo apt-get update
```
tych0, line 21: Is there some way to reference all this code so we don't have two copies of the same thing? Or perhaps only run the automatic releases bit with a condition?

tych0: Your first solution seems reasonable to me.

peusebiu:
Ok, I found that we can reuse workflows.
I created a reusable build stacker which caches it's binary workflow which I reference in ci and release.

tych0:
Cool, looks great to me, thanks! I'll merge this and wait for the other two PRs to pass CI, then we can push a tag and see if it works end to end :)

PR #285, opened 2022-05-04
Added a flag to specify OCI annotation namespace

peusebiu:
added --annotations-namespace arg

Signed-off-by: Petu Eusebiu <peusebiu@cisco.com>

### DIFF
--- a/api.go
+++ b/api.go
@@ -1,6 +1,16 @@
 package stacker
 
+import "fmt"
+
 const (
-	GitVersionAnnotation      = "com.cisco.stacker.git_version"
-	StackerContentsAnnotation = "com.cisco.stacker.stacker_yaml"
+	GitVersionAnnotation      = "%s.stacker.git_version"
+	StackerContentsAnnotation = "%s.stacker.stacker_yaml"
 )
+
+func getGitVersionAnnotation(namespace string) string {
+	return fmt.Sprintf(GitVersionAnnotation, namespace)
+}
+
+func getStackerContentsAnnotation(namespace string) string {
+	return fmt.Sprintf(StackerContentsAnnotation, namespace)
+}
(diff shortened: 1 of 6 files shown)


api.go
```diff
@@ -1,6 +1,16 @@
 package stacker
 
+import "fmt"
+
 const (
-	GitVersionAnnotation      = "com.cisco.stacker.git_version"
-	StackerContentsAnnotation = "com.cisco.stacker.stacker_yaml"
+	GitVersionAnnotation      = "com.%s.stacker.git_version"
```
rchincha, line 6: remove the "com." as well

For example, this would "io.zotregistry.stacker.git_version"

rchincha: Thinking about this some more. Do we want to just default this to "io.project-stacker.stacker.git_version"

rchincha: "io.stackeroci.stacker.git_version"

peusebiu: ok, sure

squashfs/verity.go
```diff
@@ -81,7 +81,7 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-const VerityRootHashAnnotation = "com.cisco.stacker.squashfs_verity_root_hash"
+const VerityRootHashAnnotation = "io.stackeroci.stacker.squashfs_verity_root_hash"
```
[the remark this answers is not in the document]
andaaron: @hallyn what do you think?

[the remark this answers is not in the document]
hallyn: It probably shouldn't be configurable, as that will make images from different stacker builds non-interoperable, right?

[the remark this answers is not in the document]
andaaron: Ok, let's mark this as resolved.

PR #294, opened 2022-05-20
Avoid too long utsnames

hallyn:
When stacker builds a container, it uses the s.TemporaryWritableSnapshot(name)
results.  This can be > 63 chars, which lxc will balk at.  For instance:

error: failed setting config lxc.uts.name to temp-snapshot-build-cluster-api-provider-sveltos-amd64-2544038502: setting config item for the container failed

We don't actually need a good name for the container, it just needs to run.
So just use "stacker".

If we care, we could instead do something only if len(name) > 63...


andaaron (approved):

andaaron:
@mikemccracken shall we merge this as fix for https://github.com/project-stacker/stacker/issues/292?

PR #297, opened 2022-05-30
Bats test check file created in previous layer with overlayFS

chofnar:
Signed-off-by: Catalin Hofnar <catalin.hofnar@gmail.com>

### DIFF
--- a/test/overlay-dirs.bats
+++ b/test/overlay-dirs.bats
@@ -53,6 +53,36 @@ EOF
     stacker build
 }
 
+@test "build binary and import in distroless" {
+    mkdir dir_to_overlay
+    chmod -R 777 dir_to_overlay
+    cat > stacker.yaml << EOF
+build:
+    from:
+        type: oci
+        url: $CENTOS_OCI
+    binds:
+        - dir_to_overlay -> /dir_to_overlay
+    run: |
+        touch /dir_to_overlay/binaryfile
+    build_only: true
+contents:
+    from:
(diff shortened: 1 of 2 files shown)


test/overlay-dirs.bats
```diff
@@ -53,6 +53,32 @@ EOF
...
+justbinary:
+    from:
+        type: docker
+        url: docker://gcr.io/distroless/base
+    overlay_dirs:
+        - source: dir_to_overlay
+          dest: /dir_to_overlay
+EOF
+    stacker build
+    ls dir_to_overlay | grep "helloworld.txt"
```
andaaron, line 78: What we want to verify is the `justbinary` image contains `helloworld.txt`.
Please unpack `justbinary` using skopeo and verify the file is there.

andaaron (changes requested):
Please also add an example with:
```
from:
    type: scratch
```.

Also update this doc with a `scratch` example as an alternative to the other 2.
https://github.com/project-stacker/stacker/blob/master/doc/tricks.md

andaaron:
Updated my previous comment since type;scratch was dropped in https://github.com/project-stacker/stacker/commit/97a80cda7f5b500ad22834ce04a647c219e109c5

test/overlay-dirs.bats
chofnar: Added, + documentation

andaaron (approved):

Issue #301, opened 2022-06-15
error while unmarshalling stacker file (binds field)


shimish2:
If I add a binds field in stacker file like this 

``` 
binds:
        - Source: test 
          Dest: test
```

I get this error 

```error: unknown run array type: map[interface {}]interface {}``` 

but if I use binds field like this it works fine

``` 
binds:
        - test -> test 
```

darshilgit:
Seems like [this part of code](https://github.com/project-stacker/stacker/blob/e8e0df5c2cbed4ed3112c71b0f248caa0137a558/types/layer.go#L151) is assuming that it will be a string but it can be a struct with `Source` and `Dest` strings (basically of type [Bind](https://github.com/project-stacker/stacker/blob/e8e0df5c2cbed4ed3112c71b0f248caa0137a558/types/layer.go#L133)

PR #302, opened 2022-06-15
fix(bind): interpret bind defined as type []map[string]string 

darshilgit:
As per the type definition of Layer, bind is of type Bind which is a
struct with fields Source and Dest.

But when we do UnmarshalYAML for type Bind, we are assuming that we
would always get the Bind as a []string and not
[]map[interface{}]interface{} (when making a call to
getStringOrStringSlice).

This PR fixes the getStringOrStringSlice to consider Bind defined
as a []map[string]string and returns appropriate bind string
in that case.

Fixes https://github.com/project-stacker/stacker/issues/301

### DIFF
--- a/types/layer.go
+++ b/types/layer.go
@@ -44,9 +44,43 @@ type OverlayDir struct {
 
 type Imports []Import
 
-func getStringOrStringSlice(iface interface{}, xform func(string) ([]string, error)) ([]string, error) {
+func validateDataAsBind(i interface{}) (map[interface{}]interface{}, error) {
+	bindMap, ok := i.(map[interface{}]interface{})
+	if !ok {
+		return nil, errors.Errorf("unable to cast into map[interface{}]interface{}: %T", i)
+	}
+
+	// validations
+	bindSource, ok := bindMap["Source"]
+	if !ok {
+		return nil, errors.Errorf("bind source missing: %v", i)
+	}
+
+	_, ok = bindSource.(string)
+	if !ok {
+		return nil, errors.Errorf("unknown bind source type, expected string: %T", i)
+	}
+
+	bindDest, ok := bindMap["Dest"]
+	if !ok {
+		return nil, errors.Errorf("bind dest missing: %v", i)
+	}
+
+	_, ok = bindDest.(string)
+	if !ok {
+		return nil, errors.Errorf("unknown bind dest type, expected string: %T", i)
+	}
+
+	if bindSource == "" || bindDest == "" {
+		return nil, errors.Errorf("empty source or dest: %v", i)
+	}
+
+	return bindMap, nil
+}
+
+func getStringOrStringSlice(data interface{}, xform func(string) ([]string, error)) ([]string, error) {
 	// The user didn't supply run: at all, so let's not do anything.
-	if iface == nil {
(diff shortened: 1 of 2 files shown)


mikemccracken:
LGTM, let's add a test case for this

test/caching.bats
```diff
@@ -163,6 +163,36 @@ EOF
...
 }
 
+@test "bind as struct" {
+    cat > stacker.yaml <<"EOF"
+bind-test:
+    from:
+        type: oci
+        url: ${{CENTOS_OCI}}
+    import:
+        - tree1/foo/bar
```
mikemccracken, line 173: I think this can be simplified. we don't need an import of tree1 to test binding tree2. I think the approach we want is to bind in some path, then make a change (add a file, say) in the stacker `run` section, then after the stacker run completes, we check if that new file exists at the path we bound in. we only need one path, not both tree1 and tree2

test/caching.bats
```diff
@@ -163,6 +163,36 @@ EOF
...
+    echo foo >> tree1/foo/bar
+    mkdir -p tree2/foo
+    echo bar >> tree2/foo/bar
+
+    bind_path=$(realpath tree2/foo)
+
+    # The layer should be built
+    stacker build --substitute bind_path=${bind_path} --substitute CENTOS_OCI=$CENTOS_OCI
+    out=$(stacker build --substitute bind_path=${bind_path} --substitute CENTOS_OCI=$CENTOS_OCI)
+    [[ "${out}" =~ ^(.*filesystem bind-test built successfully)$ ]]
```
mikemccracken, line 193: we do want a check for successful builds, but line 191 and 192 are running the build twice. remove 191.

mikemccracken (changes requested):
Since we're just adding a simple syntax test that the struct syntax for binds works, how about moving this to a new file `test/binds.bats`, and add a simplified test for the old syntax too?

mikemccracken (approved):

hallyn:
@darshilgit  can you rebase and repush to retrigger build?  (the build logs have been purged)

rchamarthy (approved):

rchamarthy (approved):

rchincha:
I might this in PR https://github.com/project-stacker/stacker/pull/349

hallyn (approved):
Thank you

rchamarthy (approved):

types/layer.go
```diff
@@ -58,8 +58,44 @@ func getStringOrStringSlice(iface interface{}, xform func(string) ([]string, err
...
+			case string:
+				s = v
+			case interface{}:
+				bindMap, ok := v.(map[interface{}]interface{})
+				if !ok {
+					return nil, errors.Errorf("unable to cast into map[interface{}]interface{}: %T", i)
+				}
+
+				// validations
+				bindSource, ok := bindMap["Source"]
```
shimish2, line 72: Can we have common method to check for key and cast it to string ?

hallyn (approved):

shimish2 (approved):

rchincha:
Not particularly happy with that we are turning golang to loosey-goosey javascript here, but a cleanup for another day.

PR #307, opened 2022-09-20
Fix bug in overlay-dirs cache test

peusebiu:
Test is assumming the map where modified overlay-dirs is sorted.

Signed-off-by: Petu Eusebiu <peusebiu@cisco.com>


andaaron (approved):

rchincha:
Let's keep the assertions but with adjustments.

andaaron (approved):

Issue #41, opened 2019-07-15
RFE: Allow import to take destination path


smoser:
WIP: allow import to take destination path.

When using --substitute with a stacker file, I came across a stacker the desire to have a destination path on the import.

I had a stacker file that was using --substitute to get a directory 'import'ed.  But then we wanted to use the path to that directory inside the 'run' section.

Imagine that the default 'SCRIPTS_DIR' below is named '/path/to/scripts', and is expected to have a script named 'do-magic' in it.

    import:
     - ${{SCRIPTS_DIR}}
    run: |
      PATH=/stacker/scripts:$PATH
      echo "running your script"
      do-magic

That will all work well.

However, if we want to change the name of SCRIPTS_DIR to 'common-scripts' then we'd have to change the yaml in two otherwise unrelated places (import and /run).  It is also unclear from reading this yaml file where /scripts comes from at all.  It seems annoying at least that there is an expectation on the content of the 'SCRIPTS_DIR' variable (it must end in /scripts).

What I'm proposing for a fix here is instead:

    import:
      - ${{SCRIPTS_DIR}} -> /scripts

Then the /scripts string is present in the two places that it is used and updates to the value of SCRIPTS_DIR do not force changesto the 'run:' section.

mikemccracken:
How about having the import entry be a dict?
```yaml
import:
    - /a/regular/file
    - { source: "/a/badly/named/f1l3", dest: "/a/better/name" }
```
Seems a little more in keeping with the rest of stacker, ie using yaml instead of making new syntax…

tych0:
The only thing is I already fucked up and invented this new syntax for "binds", it's just that we don't really use it.

smoser:
Right. I just followed existing implementation on binds.

There dictionary does make sense, though.

tych0:
I'd be open to a dictionary, and just dropping the other sytnax support from binds too.

smoser:
one thing that is nice with an array of strings versus a dictionary is the human friendly-ness. example:

    import:
      - http://your.example.com/my.tar.gz -> /download.tar.gz
      - my-local-dir/

versus:

    import:
      - {'source': 'http://your.example.com/my.tar.gz', 'dest': '/download.tar.gz'}
      - {'source': 'my-local-dir/'}

or alternatively this is less extensible and "forces" declaration of the target:

    import:
      /download.tar.gz: http://your.example.com/my.tar.gz
      /my-local-dir: my-local-dir

[...]

PR #320, opened 2022-10-26
feat(copy): add a copy directive

rchincha:

### DIFF
--- a/test/copy.bats
+++ b/test/copy.bats
@@ -1,0 +1,44 @@
+load helpers
+
+function setup() {
+    stacker_setup
+}
+
+function teardown() {
+    cleanup
+}
+
+@test "built type layers are restored correctly" {
+    cat > stacker.yaml <<EOF
+parent:
+    from:
+        type: oci
+        url: $CENTOS_OCI
+    run: |
+        touch /root/parent
+        cat /proc/self/mountinfo
+child:
+    from:
+        type: built
+        tag: parent
+    copy:
+      - source: stacker://parent/root/parent
+        dest: /root/parent
+        perms: 0744
+        user: user
+        group: group
+    run: |
+        cat /proc/self/mountinfo
+        [ -f /root/parent ]
+        touch /root/child
+EOF
+    stacker build
+
+    umoci --log=debug unpack --image oci:parent dest/parent
+    [ "$status" -eq 0 ]
+    [ -f dest/parent/rootfs/root/parent ]
+
+    umoci --log info unpack --image oci:child dest/child  # say my name say my name
+    [ "$status" -eq 0 ]
+    [ -f dest/child/rootfs/root/child ]
+}
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ BUILD_TAGS = exclude_graphdriver_btrfs exclude_graphdriver_devicemapper containe
 
 STACKER_OPTS=--oci-dir=.build/oci --roots-dir=.build/roots --stacker-dir=.build/stacker --storage-type=overlay
 
-build_stacker = go build -tags "$(BUILD_TAGS) $1" -ldflags "-X main.version=$(VERSION_FULL) -X main.lxc_version=$(LXC_VERSION) $2" -o $3 ./cmd
+build_stacker = go build -tags "$(BUILD_TAGS) $1" -ldflags "-X main.version=$(VERSION_FULL) -X main.lxc_version=$(LXC_VERSION) $2" -gcflags all='-N -l' -o $3 ./cmd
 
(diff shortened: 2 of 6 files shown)


test/copy.bats
```diff
@@ -0,0 +1,40 @@
...
+        [ -f /root/parent ]
+        touch /root/child
+EOF
+    stacker build
+
+    umoci --log=debug unpack --image oci:parent dest/parent
+    [ "$status" -eq 0 ]
+    [ -f dest/parent/rootfs/root/parent ]
+
+    umoci --log info unpack --image oci:child dest/child  # say my name say my name
```
mikemccracken, line 37: 👏

test/copy.bats
```diff
@@ -0,0 +1,40 @@
...
+        url: $CENTOS_OCI
+    run: |
+        touch /root/parent
+        cat /proc/self/mountinfo
+child:
+    from:
+        type: built
+        tag: parent
+    copy:
+      - /stacker/root/parent -> /root/parent
```
mikemccracken, line 25: I don't follow this - how did /stacker/root/parent get there? it wasn't imported.
/root/parent should already be there when the child container is started because it is  in the parent rfs, right?

IIUC the use case for copy is to copy something from outside the container in, as in dockerfile's `COPY`.
For that use case we could also consider e.g. adding a `destination` field to `import` directives:

```yaml
import:
  - path: foo.conf
    hash: <optional>
    dest: /etc/foo/foo.conf
```

hallyn, line 25: Is there a clean way we can augment this to optionally specify uid, gid, and perms?

Something like

```
copy:
  - /stacker/bar -> /root/bar
  - /stacker/foo -> /usr/bin/foo 0:daemon 0755
  - /stacker/baz -> /etc/baz 0:0 644
  - ```

Comment (rchincha) [reply test/copy.bats:25]: The solution for this would be an intermediate layer which fixes the perms before "copy"ing the file.
We can do this, but a minimal directive preferable?

Comment (hallyn) [reply test/copy.bats:25]: For copying files which were built by some other layer, that makes sense (though it still may be comforting to see the final owner+perms in the final stackerfile, but that's somewhat subjective).  But if I'm shipping a git tree that just contains two files to copy, and a stackfile which just creates an empty oci image with just those two files (for instance, a manifest and manifest signature), from: empty, then you're making me download some sort of distro image just to build an intermediate layer in which to import those two files, chown them, and then let the final layer export them.

So IMO if we're going to support copy, we may as well make it as easy as possible to use...

File: test/copy.bats [hunk 3]
```diff
@@ -0,0 +1,40 @@
...
+        url: $CENTOS_OCI
+    run: |
+        touch /root/parent
+        cat /proc/self/mountinfo
+child:
+    from:
+        type: built
+        tag: parent
+    copy:
+      - stacker://parent/root/parent -> /root/parent
```
smoser, line 25: I thought I saw @mikemccracken complain elsewhere about the ascii art here.
I'm not fan.  We have a reasonable data format, why not use it?

```yaml
copy:
 - { src: "stacker://parent/root/parent", target: "/root/parent", perms: 0755, owner: "1001" }
 - ...
```
The choice to use ascii art bit us https://github.com/project-stacker/stacker/issues/41 also.

rchincha: makes the syntax a little verbose, but unambiguous

mikemccracken: one thought I had about this is that now people have to remember which of `src`, `source`, `dst`, `dest`, `destination`, `tgt`, `target`, etc are the right keys.

My suggestion is to just allow all of them.

smoser: Generally speaking I have convinced myself that "Be liberal in what you accept, and conservative in what you send" is wrong.  It just gets you into a support issue.  Its confusing for users too, they see some examples that say 'src' and others that say 'source' and they don't know which is "right" (which even *is* right?). And you can't fix the situation without breaking a user.

The best thing to do is be consistent and give good errors:  unknown key 'src' in copy element 1.  Did you mean 'source' ?


That said, I do think allowing a string "0755" is helpful, compared to only allowing that to be an integer.  That said, the following actually is correct. yaml reads 0755 as base8.

```
$ echo 'myperms: 0755' | python3 -c 'import sys, yaml; print(yaml.safe_load(sys.stdin)["myperms"])'
493
```

smoser: I realize... its funny.  *yaml* is being very liberal in what it accepts , and thats a win for us ;). Yaml being liberal though and "friendly" to its users also ends up confusing them lots of times, which I'm sure we've all ran into.

mikemccracken: OK, yeah that's sane. Good errors solves that problem in a better way.

PR #322, opened 2022-11-10
Drop overlay attributes before creating squashfs.

smoser:
The overlay attributes [1] have given us some grief [2]. They don't seem to provide benefit.  We are keeping the overlay.opaque values, but drop everything else in *.overlay.* namespace.

[1] https://www.kernel.org/doc/html/latest/filesystems/overlayfs.html 
[2] https://github.com/vasi/squashfuse/issues/78

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,6 +30,7 @@ require (
 	github.com/opencontainers/runc v1.1.2 // indirect
 	github.com/opencontainers/umoci v0.4.8-0.20220412065115-12453f247749
 	github.com/pkg/errors v0.9.1
+	github.com/pkg/xattr v0.4.9
 	github.com/prometheus/common v0.32.1 // indirect
 	github.com/smartystreets/assertions v1.0.1 // indirect
 	github.com/smartystreets/goconvey v1.6.4
@@ -41,7 +42,7 @@ require (
 	go.mozilla.org/pkcs7 v0.0.0-20210826202110-33d05740a352 // indirect
 	golang.org/x/crypto v0.0.0-20211117183948-ae814b36b871 // indirect
 	golang.org/x/net v0.0.0-20220225172249-27dd8689420f // indirect
-	golang.org/x/sys v0.0.0-20220227234510-4e6760a101f9
+	golang.org/x/sys v0.0.0-20220408201424-a24fb2fb8a0f
 	golang.org/x/term v0.0.0-20210927222741-03fcf44c2211
 	google.golang.org/genproto v0.0.0-20220303160752-862486edd9cc // indirect
 	gopkg.in/yaml.v2 v2.4.0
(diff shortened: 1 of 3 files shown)


hallyn:
Should the trusted.overlay.opaque xattrs be renamed to user.overlay.opaque?

smoser:
> Should the trusted.overlay.opaque xattrs be renamed to user.overlay.opaque?

It feels to me like that should be a *mount* time thing, rather than a fs-create time thing.  The only thing that I see about 'trusted.overlay' vs 'user.overlay' namespace is from [kernel overlay doc](https://www.kernel.org/doc/html/latest/filesystems/overlayfs.html) .:

> The “-o userxattr” mount option forces overlayfs to use the “user.overlay.” xattr namespace instead of “trusted.overlay.”. This is useful for unprivileged mounting of overlayfs.

What do you think, am I missing something?  Suggesting that I should modify the creation of a filesystem to account for unpriviledged mounts seems somewhat like suggesting I should shift all my UIDs by 65535 so that a usernamespace would have the right uid/gid mappings. it seems not the right place to do it.

It is absolutely p ossible I'm missing something.

hallyn:
So actually we shouldn't need to do a post-build step, we should be able to just mount over build overlays with the ,userxattr option.

Regarding a mount option - how would it work?  Would it be an overlay mount option which says "when copying up a file that has a trusted.overlay xattr, ignore that xattr"?

hallyn:
I opened https://github.com/project-stacker/stacker/pull/323  - I'm fuzzy on whether that would be a replacement or an enhancement to this PR.  I'm thinking replacement right now.  What do you think?

Confirmed with a toy busybox based local testcase that i now get user xattr:
```
$ ~/test/llistxattrfull /mnt/a
ret 20 on /mnt/a
nret 20 list user.overlay.origin
```

smoser:
> So actually we shouldn't need to do a post-build step, we should be able to just mount over build overlays with the ,userxattr option.
> 
> Regarding a mount option - how would it work? Would it be an overlay mount option which says "when copying up a file that has a trusted.overlay xattr, ignore that xattr"?

Isn't that what "-o userxattr does" ?

hallyn:
> > So actually we shouldn't need to do a post-build step, we should be able to just mount over build overlays with the ,userxattr option.
> > Regarding a mount option - how would it work? Would it be an overlay mount option which says "when copying up a file that has a trusted.overlay xattr, ignore that xattr"?
> 
> Isn't that what "-o userxattr does" ?

No.  -o userxattr will create new xattrs are user.*, but I don't believe it will prevent breakage if you try to copy up a file from an underlaying fs which has a trusted.overlay.* xattr.

hallyn:
So, do we still want this patch, @smoser ?  (Last night I was thinking no, but now I'm thinking yes.)

hallyn (approved):

hallyn (approved):

PR #323, opened 2022-11-15
Mount overlay with userxattr option

hallyn:
This will cause all upperdirs to user user.overlay.* instead of trusted.overlay.* xattrs.  Since a root user may generate a layer which will later be used as a base by an unprivileged build, there's no reason to ever use trusted.*.

### DIFF
--- a/overlay/overlay.go
+++ b/overlay/overlay.go
@@ -218,7 +218,7 @@ func (o *overlay) GetLXCRootfsConfig(name string) (string, error) {
 		return "", err
 	}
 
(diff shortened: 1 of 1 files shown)


hallyn:
With this, layers get created with user xattrs instead of trusted.*.

```
$ ~/test/llistxattrfull /mnt/a
ret 20 on /mnt/a
nret 20 list user.overlay.origin
```

smoser (commented):
This looks right to me.

hallyn:
@rchincha or @andaaron  if you have concerns with this please let me know, else I'll merge it shortly.

PR #325, opened 2022-11-16
check: update stacker check to verify newuidmap

### LINKED ISSUE #324: 'stacker check' does not check newuidmap privs
If newuidmap is not setuid-root and has no filecaps, 'stacker check' does not warn about that.  It should do so, at least when called by a non-root user, probably always.

rchamarthy:
stacker check command now verifies if the newuidmap binary exists in PATH and if the binary is executable with setuid bit on.

Fixes #324

Signed-off-by: Ravi Chamarthy <ravi@chamarthy.dev>

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -1,9 +1,12 @@
 package main
 
 import (
+	"io/fs"
 	"os"
+	"os/exec"
 
 	"github.com/pkg/errors"
+	"github.com/pkg/xattr"
 	"github.com/project-stacker/stacker/overlay"
 	"github.com/urfave/cli"
 )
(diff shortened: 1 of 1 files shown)


raharper (commented):
What do you think about have check emit suggested fixes?

cmd/check.go
```diff
@@ -25,4 +27,28 @@ func doCheck(ctx *cli.Context) error {
 	default:
 		return errors.Errorf("invalid storage type %v", config.StorageType)
 	}
+
+	return verifyNewUIDMap(ctx)
```
hallyn, line 31: Thanks for doing this.

Just move this above the switch(), and make it "if err := verifynewUIDMap(ctx); err != nil { return err };

cmd/check.go
```diff
@@ -25,4 +27,28 @@ func doCheck(ctx *cli.Context) error {
...
+	fileInfo, err := os.Stat(binFile)
+	if err != nil {
+		return errors.Wrapf(err, "couldn't stat file: %s", binFile)
+	}
+
+	if fileInfo.Mode()&0100 == 0 {
+		return errors.Errorf("%s is not executable", binFile)
+	}
+
+	if fileInfo.Mode()&fs.ModeSetuid == 0 {
```
hallyn, line 49: There are distros where newuidmap is not setuid-root, but has the CAP_SYS_ADMIN and CAP_SETUID filecaps instead.  Here things become murky, though, because the file could just have the inheritable and effective caps, and the host give permitted users the inheritable process caps,   But I doubt anyone does that, so just checking for permitted and effective is probably best.

rchamarthy:
> What do you think about have check emit suggested fixes?

It's a good idea, but I don't know if we can suggest the correct fix that will work consistently with the distro.

rchamarthy:
> What do you think about have check emit suggested fixes?

It's a good idea, but I don't know if we can suggest the correct fix that will work consistently with the distro.

hallyn:
> > What do you think about have check emit suggested fixes?
> 
> It's a good idea, but I don't know if we can suggest the correct fix that will work consistently with the distro.

Agreed.  For ubuntu and fedora the fixes are quite different.  Well, setuid-root will always work, but...

rchincha:
Can we just detect the distro and suggest?
https://github.com/zcalusic/sysinfo, for example.

rchamarthy:
> Can we just detect the distro and suggest? https://github.com/zcalusic/sysinfo, for example.

I think this is a good start, after this its googling the error :)

hallyn (approved):
Thanks

rchincha (approved):
lgtm

PR #327, opened 2022-11-18
chore: upgrade dependencies (lxc, ubuntu, golang) 

rchincha:


Makefile
```diff
@@ -30,7 +30,7 @@ stacker-dynamic: $(GO_SRC) go.mod go.sum cmd/lxc-wrapper/lxc-wrapper
 	$(call build_stacker,,,stacker-dynamic)
 
 cmd/lxc-wrapper/lxc-wrapper: cmd/lxc-wrapper/lxc-wrapper.c
-	make -C cmd/lxc-wrapper LDFLAGS=-static LDLIBS="$(shell pkg-config --static --libs lxc) -lpthread -ldl" lxc-wrapper
+	make -C cmd/lxc-wrapper LDFLAGS=-static CFLAGS="-I/usr/include" LDLIBS="$(shell pkg-config --static --libs lxc) -lpthread -ldl -llxc" lxc-wrapper
```
smoser, line 33: Surprised you need -I/use/include. Are you sure?

rchincha:
https://github.com/lxc/lxc/issues/4223
https://github.com/lxc/lxc/pull/4224

rchamarthy (approved):
lgtm

build.yaml
```diff
@@ -55,11 +55,18 @@ build-env:
...
+    apk add meson ninja docbook2x docbook2x-doc curl
     git clone --depth 1 ${{LXC_CLONE_URL}} -b ${{LXC_BRANCH}}
     cd lxc
-    ./autogen.sh
-    ./configure --enable-static-binaries --prefix=/usr
-    make -j$(grep -c processor /proc/cpuinfo) install
+    mkdir /etc/default
+    # patches
+    ## static lib name
+    curl -L https://github.com/lxc/lxc/commit/64eb31d02d4933ad414239130df3dff45a0f6f91.diff | git apply
```
smoser, line 64: This looks odd.  
Something like it would be necessary if we were building a static commit, but since we're trackign a branch, I put up https://github.com/lxc/lxc/pull/4225 to get that into the stable-5.0 branch.

So, if that lands, it will actually break builds here.

build.yaml
```diff
@@ -55,11 +55,18 @@ build-env:
...
+    ## date check
+    sed -i '311s/$/.strip()/' meson.build
```
smoser, line 66: does this have an upstream issue or pr associated with it? Just reading, I have no context on why this is here.
Is this the change you were actually intending to make?

```
$ git checkout upstream/stable-5.0 
HEAD is now at 116c24ade Merge pull request #4215 from cyphar/lxc5.x-opensuse-build
$ sed -i '311s/$/.strip()/' meson.build
$ git diff
diff --git a/meson.build b/meson.build
index 93572975e..59c386bec 100644
--- a/meson.build
+++ b/meson.build
@@ -308,7 +308,7 @@ if not want_sd_bus.disabled()
             oss_fuzz_dependencies += libsystemd
         endif
     endif
-
+.strip()
     srcconf.set10('HAVE_LIBSYSTEMD', has_sd_bus)
 else
     has_sd_bus = false
```

smoser (changes requested):
More just a curiosity.
Why did you bump the ubuntu version?
Why did you bump the go version? ( and drop 1.17 and 1.18)?

Neither one of those things are obviously necessary.  I'm guessing the go version bump was required, but it is good to include things like that in commit messages.

I hope the review is useful, I'm not trying to be pedantic, but instead looking at the changes as I would be reading them a couple months down the road.

build.yaml
smoser: perhaps its related to https://github.com/lxc/lxc/pull/4224 . thanks for filing an upstream issue and PR, please put a reference in here if that is what you're fixing.

rchincha:
> More just a curiosity. Why did you bump the ubuntu version? Why did you bump the go version? ( and drop 1.17 and 1.18)?
> 
> Neither one of those things are obviously necessary. I'm guessing the go version bump was required, but it is good to include things like that in commit messages.
> 
> I hope the review is useful, I'm not trying to be pedantic, but instead looking at the changes as I would be reading them a couple months down the road.

I will update the commit msg. Trying to update everything in one-shot so we don't have to worry about tech debt for a little while.

hallyn (approved):

PR #328, opened 2022-11-18
stacker_yaml doc: fix indenting

mikemccracken:
the section header depths made the outline weird, they should match the levels in the actual yaml, except for `import hash` which seemed enough like a sub-type that I left it as a child of `import`.

Signed-off-by: Michael McCracken <mikmccra@cisco.com>


mikemccracken:
image
what i meant by the outline being weird

rchincha (approved):
lgtm

Issue #319, opened 2022-10-21
Unable to use squashfs without tar layers


smoser:
I'm trying to use only squashfs layers, and running into a problem that occurs both as unprivileged and privileged user.

Basically trying to build this stacker.yaml with only `--layer-type=squashfs`
```yaml
rootfs:
  from:
    type: docker
    url: "${{BASE_URL}}${{LAYER_SUFFIX}}"
  import:
    - hello-world
  run: |
    cp /stacker/hello-world /usr/bin/
    chmod 755 /usr/bin/hello-world
    /usr/bin/hello-world
```

This can be reproduced with [go-test.sh](https://github.com/project-stacker/stacker/files/9842697/go-test.sh.txt) as root or non-root.  Just point it at a url that has both -squashfs and tar formats.  Both root and non-root paths fail.

Output of go-test.sh is attached:
 * as root : [root.log](https://github.com/project-stacker/stacker/files/9842702/root.log)
 * as non-root [no-root.log](https://github.com/project-stacker/stacker/files/9842700/no-root.log)

smoser:
I've just now tested with stacker 0.3.1 and it fails as both root and non-root.

[root-0.3.1.log.txt](https://github.com/project-stacker/stacker/files/10127562/root-0.3.1.log.txt) `sudo rm -Rf sdir-*; sudo PATH=/tmp/stackers/0.30.1:$PATH ./go-test2.sh  2>&1 | tee no-root-0.3.1.log`
[no-root-0.3.1.log.txt](https://github.com/project-stacker/stacker/files/10127563/no-root-0.3.1.log.txt) `sudo rm -Rf sdir-*; PATH=/tmp/stackers/0.30.1:$PATH ./go-test2.sh  2>&1 | tee no-root-0.3.1.log`

The one change to the script is setting the baseurl (to something built with a recent stacker).


#329 fixes this issue.

See (v0.22.1-8-gb287f83)
[root-dev.log](https://github.com/project-stacker/stacker/files/10127583/root-dev.log) - `sudo rm -Rf sdir-*; sudo PATH=/tmp/stackers/dev:$PATH ./go-test2.sh  2>&1 | tee /tmp/root-dev.log`
[no-root-dev.log](https://github.com/project-stacker/stacker/files/10127584/no-root-dev.log) - `sudo rm -Rf sdir-*; PATH=/tmp/stackers/dev:$PATH ./go-test2.sh  2>&1 | tee /tmp/no-root-dev.log`

[...]

PR #329, opened 2022-11-22
Feature/fuse maybe

smoser:
Lots of debug here.
definitely work in progress. locally it passes a few small tests for me.

### DIFF
--- a/cmd/lxc-wrapper/lxc-wrapper.c
+++ b/cmd/lxc-wrapper/lxc-wrapper.c
@@ -6,11 +6,20 @@
 #include <signal.h>
 #include <sched.h>
 #include <errno.h>
+#include <sys/mount.h>
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <sys/socket.h>
 
 #include <lxc/lxccontainer.h>
+
+#define STACK_SIZE (1024 * 1024)
+
+struct child_args {
+	char **argv;          /* Command to be executed by child, with args */
+	int    sk_pair[2];    /* Socket used to synchronize parent and child */
+	int    command_start;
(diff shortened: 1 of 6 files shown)


rchincha:
Would squash it once done.

smoser:
> Would squash it once done.

I think it is done. I'd prefer to leave it as 4 commits. it is 4 units of work.

hallyn (approved):

cmd/lxc-wrapper/lxc-wrapper.c
```diff
@@ -71,102 +80,153 @@ static int map_id(char *args[])
...
+	// The intent is that our child process is in:
+	//  * user namespace for 'usernsexec' (CLONE_NEWUSER)
+	//  * private (MS_PRIVATE) mount namespace always (CLONE_NEWNS)
+	//  * pid namespace always (CLONE_NEWPID)
+	//
+	// If not in its' own user namespace, then mounts will propagate.
+	// change that behavior so it is consistent for both cases.
+	ret = mount("none", "/", 0, MS_PRIVATE | MS_REC, NULL);
+	if (ret != 0) {
+		fprintf(stderr, "mounting proc failed: %m\n");
```
smoser, line 99: Ugh. Mounting proc failed. Should change that message.

smoser:
I guess one thing that I'd really like to have added.... but I'm not wanting to do it right now, is a test case like shown in #319.

These changes fix that test case (as I commented in the issue).

Basically, before this commit, we could not do "squashfs only" builds either privileged or unprivileged without starting
`from` a `type: docerk` `url` that referenced tar content.

PR #331, opened 2022-11-22
Check subuid setup and such

smoser:
testing, one two testing.


smoser:
testing only. closing.

PR #333, opened 2022-11-28
Explicitly avoid trying to remove xattrs on symlink

hallyn:
You cannot have user.* xattrs on a symlink.

Except...  overlay is in the kernel, it can do what it wants. Unfortunately, fs/xattr.c:xattr_permission() won't allow us to do anything with them:

 124         /*
 125          * In the user.* namespace, only regular files and directories can have
 126          * extended attributes. For sticky directories, only the owner and
 127          * privileged users can write attributes.
 128          */
 129         if (!strncmp(name, XATTR_USER_PREFIX, XATTR_USER_PREFIX_LEN)) {
 130                 if (!S_ISREG(inode->i_mode) && !S_ISDIR(inode->i_mode))
 131                         return (mask & MAY_WRITE) ? -EPERM : -ENODATA;

So just skip trying to remove any xattrs on symlinks.

This *should* be ok, because if I do mksquashfs of a directory with a symlink with user.overlay.origin xattr and then use squashfuse to mount it, the xattr is no longer there.

### DIFF
--- a/lib/dir.go
+++ b/lib/dir.go
@@ -7,6 +7,16 @@ import (
 	"github.com/pkg/errors"
 )
 
+func IsSymlink(p string) bool {
+	fi, err := os.Lstat(p)
+	if err != nil {
+		// Some people can't be helped
+		return false
+	}
+
+	return fi.Mode()&os.ModeSymlink != 0
+}
+
 // DirCopy copies a whole directory recursively
 func DirCopy(dest string, source string) error {
 
--- a/overlay/pack.go
+++ b/overlay/pack.go
@@ -352,7 +352,12 @@ func stripOverlayAttrsUnder(dirPath string) error {
 			if err != nil {
 				return err
 			}
-			return stripOverlayAttrs(filepath.Join(dirPath, path))
+			p := filepath.Join(dirPath, path)
+			if lib.IsSymlink(p) {
+				// user.* xattrs "can not" exist on symlinks
+				return nil
+			}
+			return stripOverlayAttrs(p)
 		})
 }
 


hallyn:
```
Starting registry service container
  /usr/bin/docker pull registry:2
  Error response from daemon: Head "https://registry-1.docker.io/v2/library/registry/manifests/2": received unexpected HTTP status: 503 Service Unavailable
  Warning: Docker pull failed with exit code 1, back off 1.452 seconds before retry.
  /usr/bin/docker pull registry:2
  Error response from daemon: Head "https://registry-1.docker.io/v2/library/registry/manifests/2": received unexpected HTTP status: 503 Service Unavailable
  Warning: Docker pull failed with exit code 1, back off 4.541 seconds before retry.
  /usr/bin/docker pull registry:2
  Error response from daemon: Head "https://registry-1.docker.io/v2/library/registry/manifests/2": received unexpected HTTP status: 503 Service Unavailable
  Error: Docker pull failed with exit code 1
```

smoser (approved):
Thanks

hallyn:
FWIW with mainline kernel in ubuntu jammy I don't get these stray xattrs.  With 5.15.0-53-generic kernel in ubuntu jammy, I do.

smoser:
> FWIW with mainline kernel in ubuntu jammy I don't get these stray xattrs. With 5.15.0-53-generic kernel in ubuntu jammy, I do.

it would be nice to file an ubuntu bug if you can.

hallyn:
> > FWIW with mainline kernel in ubuntu jammy I don't get these stray xattrs. With 5.15.0-53-generic kernel in ubuntu jammy, I do.
> 
> it would be nice to file an ubuntu bug if you can. That’d also give sobering to point at for a fix.

I can give pretty simple reproduction instructions using stacker, but it'd be nicer if I could do so with just a manual overlay example.

smoser:
> > it would be nice to file an ubuntu bug if you can. That’d also give sobering to point at for a fix.

"sobering". Thanks iphone.  That should have said: That'd also give us something to point at for a fix.

hallyn:
It took me a few minutes but I did decipher it as 'something' :-)

hallyn:
I've failed to reproduce this by hand so far.  But using stacker, it's just this stacker.yaml:

```

ubuntu@cmpj:~/build2$ cat stacker.yaml
pxe-server-base:
    from:
        type: docker
        url: docker://ubuntu:jammy
    run: |
        apt-get update
        apt-get -y install dnsmasq systemd

sb-pxe-server:
    from:
        type: built
        tag: pxe-server-base
    run: |
      systemctl disable dnsmasq
```

which ends with

```
Executing: /lib/systemd/systemd-sysv-install disable dnsmasq
Removed /etc/systemd/system/multi-user.target.wants/dnsmasq.service.
error: /home/ubuntu/build2/roots/sb-pxe-server/overlay/etc/rc2.d/K01dnsmasq: failed to remove attr user.overlay.origin: xattr.LRemove /home/ubuntu/build2/roots/sb-pxe-server/overlay/etc/rc2.d/K01dnsmasq user.overlay.origin: operation not permitted
error: exit status 1
```

hallyn:
Maybe instead of simply ignoring all errors, we should try to read the contents.  If that returns ENODATA or EPERM then don't try to remove it.  But it's more work for probably very little gain.

rchincha (approved):
lgtm

PR #345, opened 2022-11-30
feat(import): add "copy" support for import

rchincha:
Currently, "import:" directive only allows for temporarily mounting previously layers, and actual copying would have to be done via "run:" directive. But in cases where there is no shell in the current layer, there was no way to achieve this, greatly limiting usability for packaging golang binaries.

Adding support using "dest:" option under "import:" directive.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/lib/dir.go
+++ b/lib/dir.go
@@ -1,8 +1,11 @@
 package lib
 
 import (
+	"io/fs"
 	"os"
 	"path"
+	"strconv"
+	"strings"
 
 	"github.com/pkg/errors"
 )
@@ -62,7 +65,7 @@ func DirCopy(dest string, source string) error {
 				return err
 			}
 		} else {
-			if err = FileCopy(dstfp, srcfp); err != nil {
+			if err = FileCopy(dstfp, srcfp, nil, -1, -1); err != nil {
 				return err
 			}
 		}
@@ -80,6 +83,62 @@ func CopyThing(srcpath, destpath string) error {
 	if srcInfo.IsDir() {
 		return DirCopy(destpath, srcpath)
 	} else {
-		return FileCopy(destpath, srcpath)
+		return FileCopy(destpath, srcpath, nil, -1, -1)
 	}
+}
+
+// Chmod changes file permissions
+func Chmod(mode, destpath string) error {
+	destInfo, err := os.Lstat(destpath)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	if destInfo.IsDir() {
+		return errors.WithStack(os.ErrInvalid)
+	}
+
+	if destInfo.Mode()&os.ModeSymlink != 0 {
+		return errors.WithStack(os.ErrInvalid)
+	}
+
+	// read as an octal value
+	iperms, err := strconv.ParseUint(mode, 8, 32)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	return os.Chmod(destpath, fs.FileMode(iperms))
+}
+
+// Chown changes file ownership
+func Chown(owner, destpath string) error {
+	destInfo, err := os.Lstat(destpath)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	if destInfo.IsDir() {
+		return errors.WithStack(os.ErrInvalid)
+	}
+
+	owns := strings.Split(owner, ":")
+	if len(owns) > 2 {
+		return errors.WithStack(os.ErrInvalid)
+	}
+
+	uid, err := strconv.ParseInt(owns[0], 10, 32)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	gid := int64(-1)
+	if len(owns) > 1 {
+		gid, err = strconv.ParseInt(owns[1], 10, 32)
+		if err != nil {
+			return errors.WithStack(err)
+		}
+	}
+
+	return os.Lchown(destpath, int(uid), int(gid))
 }
--- a/lib/file.go
+++ b/lib/file.go
@@ -2,6 +2,7 @@ package lib
 
 import (
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -9,7 +10,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-func FileCopy(dest string, source string) error {
+func FileCopy(dest string, source string, mode *fs.FileMode, uid, gid int) error {
 	os.RemoveAll(dest)
 
 	linkFI, err := os.Lstat(source)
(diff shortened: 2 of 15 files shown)


rchincha:
Don't merge this yet, only part 1/2 of the patchset

lib/dir.go
```diff
@@ -80,6 +83,58 @@ func CopyThing(srcpath, destpath string) error {
 	if srcInfo.IsDir() {
 		return DirCopy(destpath, srcpath)
 	} else {
-		return FileCopy(destpath, srcpath)
+		return FileCopy(destpath, srcpath, nil, nil, nil)
 	}
 }
+
+// Chmod changes file permissions
+func Chmod(perms, destpath string) error {
```
hallyn, line 91: 1. Why is setting perms on directory not supported?
2. You do however need to check for destinfo.Mode()&ModeSymlink and refuse to chmod in that case.

lib/dir.go
```diff
@@ -80,6 +83,58 @@ func CopyThing(srcpath, destpath string) error {
...
+
+	gid := int64(-1)
+	if len(owns) > 1 {
+		gid, err = strconv.ParseInt(owns[1], 10, 32)
+		if err != nil {
+			return errors.WithStack(err)
+		}
+	}
+
+	return os.Chown(destpath, int(uid), int(gid))
```
hallyn, line 139: Please use os.LChown here.

lib/file.go
```diff
@@ -44,11 +45,30 @@ func FileCopy(dest string, source string) error {
 	}
 	defer d.Close()
 
-	err = d.Chmod(fi.Mode())
+	if mode != nil {
```
hallyn, line 48: Hm, this isn't a bug in this patch per se, but i do notice that in the context up above if the copied file is a symlink, you do not chown it.  That could provide wrong results.  It would be best to LChown() it to linkFI's uid/gid.

types/layer.go
```diff
@@ -359,16 +367,42 @@ func requireImportHash(imports Imports) error {
...
 			hash = fmt.Sprintf("%v", m["hash"])
 		}
-		return Import{Hash: hash, Path: fmt.Sprintf("%v", m["path"])}, nil
+
+		if m["dest"] != nil {
+			if !filepath.IsAbs(m["dest"].(string)) {
+				return Import{}, errors.Errorf("Dest path cannot be relative for: %#v", v)
+			}
+
+			dest = fmt.Sprintf("%v", m["dest"])
```
hallyn, line 388: (i'm not sure, but) should this be dest = fmt.Sprintf("%q", m["dest"]) to ensure a properly quoted path string?

types/layer.go
```diff
@@ -379,7 +413,30 @@ func getImportFromInterface(v interface{}) (Import, error) {
...
 			hash = fmt.Sprintf("%v", m2["hash"])
 		}
-		return Import{Hash: hash, Path: fmt.Sprintf("%v", m2["Path"])}, nil
+
+		if m2["dest"] != nil {
+			if !filepath.IsAbs(m2["dest"].(string)) {
+				return Import{}, errors.Errorf("Dest path cannot be relative for: %#v", v)
+			}
+
+			dest = fmt.Sprintf("%v", m["dest"])
```
hallyn, line 422: (Same here about %q, if I'm not wrong)

types/layer.go
```diff
@@ -379,7 +413,30 @@ func getImportFromInterface(v interface{}) (Import, error) {
...
+
+		if m2["uid"] != nil {
+			uid = m2["uid"].(int)
+		}
+
+		if m2["gid"] != nil {
+			gid = m2["gid"].(int)
+		}
+
+		return Import{Hash: hash, Path: fmt.Sprintf("%v", m2["path"]), Dest: dest, Mode: &mode, Uid: &uid, Gid: &gid}, nil
```
hallyn, line 439: I'm getting a little lost in the flow here, but this doesn't seem right.  Here, and above, you return Import{} with Mode: &mode, Uid: &uid, Gid: &gid.  But at the top of this scope, mode uid and gid are simply their zero values.  I don't see where you interpret zero values as meaning "not provided".

lib/file.go
rchincha: 2022-12-13T19:42:34.4519138Z not ok 16 importing broken symlink is ok
2022-12-13T19:42:34.4519524Z # (from function `stacker' in file test/helpers.bash, line 86,
2022-12-13T19:42:34.4519909Z #  in test file test/broken-link.bats, line 24)
2022-12-13T19:42:34.4520230Z #   `stacker build' failed

lib/dir.go
```diff
@@ -80,6 +83,62 @@ func CopyThing(srcpath, destpath string) error {
...
 }
+
+// Chmod changes file permissions
+func Chmod(mode, destpath string) error {
+	destInfo, err := os.Lstat(destpath)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	if destInfo.IsDir() {
```
hallyn, line 97: Sorry, I know you said some of the fixes broke things, not sure if this was one of them - but why should chmod on a directory not be allowed?

lib/dir.go
```diff
@@ -80,6 +83,62 @@ func CopyThing(srcpath, destpath string) error {
...
+}
+
+// Chown changes file ownership
+func Chown(owner, destpath string) error {
+	destInfo, err := os.Lstat(destpath)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	if destInfo.IsDir() {
```
hallyn, line 121: Again - why should a directory not be allowed?

I won't not-approve due to these two things, but it is curious.

lib/dir.go
rchincha: What is the semantics of a directory as a target? its subdirs and files? what about their mode/ownership?

lib/file.go
```diff
@@ -24,7 +25,21 @@ func FileCopy(dest string, source string) error {
...
+		if err = os.Symlink(target, dest); err != nil {
+			return errors.Wrapf(err, "Couldn't symlink %s->%s", source, target)
+		}
+
+		if err = os.Lchown(dest, uid, gid); err != nil {
+			return errors.Wrapf(err, "Couldn't set symlink ownership %s", dest)
+		}
+
+		if mode != nil {
+			if err = os.Chmod(dest, *mode); err != nil {
```
hallyn, line 37: I think this is very wrong.  chmod on a symbolic link, which you've just created dest as, does not exist.  Instead, golang's os.Chmod does (per https://pkg.go.dev/os#Chmod) :

> Chown changes the numeric uid and gid of the named file. If the file is a symbolic link, it changes the uid and gid of the link's target

hallyn: Note you should indeed Lchown it, as you are, so that part is good.

rchincha:
```
$ ln -sfv non-existent bad-link
'bad-link' -> 'non-existent'

$ ls -al bad-link
lrwxrwxrwx 1 rchincha ifc 12 Dec 14 20:33 bad-link -> non-existent

$ chmod 755 bad-link
chmod: cannot operate on dangling symlink 'bad-link'
```

hallyn (approved):
Thank you

PR #349, opened 2022-12-05
feat: add a "stacker convert" to convert a Dockerfile

rchincha:
https://docs.docker.com/engine/reference/builder/

`
stacker help convert

 help convert
NAME:
   stacker-dynamic convert - converts a Dockerfile into a stacker yaml
file

USAGE:
   stacker-dynamic convert [command options] [arguments...]

OPTIONS:
   --Dockerfile value, -f value  the input Dockerfile (default: "Dockerfile")
`

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


rchincha:
Surprisingly, breaks here

not ok 30 bind rebuilds
# (in test file test/caching.bats, line 151)
#   `out=$(stacker build --substitute bind_path=${bind_path} --substitute CENTOS_OCI=$CENTOS_OCI)' failed

rchincha:
and not here:
ok 41 convert a Dockerfile

rchincha:
# Lookup2: name:bind-test cache:map[]
# TEST (binds): 1 [{/home/runner/work/stacker/stacker/stackertest-test_bind_rebuilds.78LQ7s/tree2/foo /root/tree2/foo}] hit: false

rchincha:
alpine/edge gcc was updated with this recently.
https://git.alpinelinux.org/aports/commit/?id=4592d806b9c8832895c32aeb53936b345b04b500

rchincha:
https://gitlab.alpinelinux.org/alpine/aports/-/issues/14537

go.mod
```diff
@@ -31,6 +32,7 @@ require (
 	golang.org/x/sys v0.3.0
 	golang.org/x/term v0.3.0
 	gopkg.in/yaml.v2 v2.4.0
+	sigs.k8s.io/yaml v1.2.0
```
hallyn, line 35: Can you explain why a second yaml library is needed?

cmd/stacker/publish.go
```diff
@@ -104,17 +108,18 @@ func doPublish(ctx *cli.Context) error {
...
-		Username:   ctx.String("username"),
-		Password:   ctx.String("password"),
-		Force:      ctx.Bool("force"),
-		Progress:   shouldShowProgress(ctx),
-		SkipTLS:    ctx.Bool("skip-tls"),
-		LayerTypes: layerTypes,
+		Config:         config,
+		ShowOnly:       ctx.Bool("show-only"),
+		Substitute:     ctx.StringSlice("substitute"),
+		SubstituteFile: ctx.String("substitute-file"),
```
hallyn, line 114: While --substitute-file is an awesome feature, it probably should have been a separate patch (so one feature can be reverted without the other if needed)

pkg/types/layer.go
```diff
@@ -502,7 +543,7 @@ func getImportFromInterface(v interface{}) (Import, error) {
 	// if it's not a map then it's a string
 	s, ok := v.(string)
 	if ok {
-		return Import{Hash: "", Path: fmt.Sprintf("%v", s)}, nil
+		return Import{Hash: "", Path: fmt.Sprintf("%v", s), Dest: "", Uid: uid, Gid: gid}, nil
```
hallyn, line 546: This is a bugfix for the import/copy pr?

hallyn (approved):
Looks good, just a few questions below.

pkg/types/layer.go
rchincha: Definitely a bug, since the "zero" value doesn't set other fields correctly.

cmd/stacker/publish.go
rchincha: substitute-file support already in stacker but for build cmd, now added to publish cmd. 
convert cmd and test cases less useful without that.
But split into separate commits now.

go.mod
rchincha: needed since needed a "better" way to do yaml mangling during conversion
https://github.com/kubernetes-sigs/yaml

pkg/types/layer.go
hallyn: Ok - ideally that would be a separate 2-line commit (in part to make later cherrypicks and reverts easier).

PR #359, opened 2022-12-13
feature: Support atomfs molecule mount in containers

hallyn:
Support mounting atomfs stacks as root in a container.

### DIFF
--- a/pkg/squashfs/squashfs.go
+++ b/pkg/squashfs/squashfs.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"strings"
 	"sync"
 	"syscall"
@@ -71,7 +72,7 @@ func (eps *ExcludePaths) AddInclude(orig string, isDir bool) {
 		}
 
 		delete(eps.exclude, p)
-		p = path.Dir(p)
+		p = filepath.Dir(p)
 	}
 
 	// now add it to the list of includes, so we don't accidentally re-add
@@ -214,6 +215,51 @@ func maybeKernelSquashMount(squashFile, extractDir string) (bool, error) {
 	return false, kernelSquashMountFailed
 }
 
+func findSquashfusePath() string {
+	if p := which("squashfuse_ll"); p != "" {
+		return p
+	}
+	return which("squashfuse")
+}
+
+var squashNotFound = errors.Errorf("squashfuse program not found")
+
+func squashFuse(squashFile, extractDir string) (*exec.Cmd, error) {
+	sqfuse := findSquashfusePath()
+	if sqfuse == "" {
+		return nil, squashNotFound
+	}
+
+	// given extractDir of path/to/some/dir[/], log to path/to/some/.dir-squashfs.log
+	extractDir = strings.TrimSuffix(extractDir, "/")
+
+	var cmdOut io.Writer
+	var err error
+	var nilCmd *exec.Cmd
+
+	logf := filepath.Join(path.Dir(extractDir), "."+filepath.Base(extractDir)+"-squashfuse.log")
+	if cmdOut, err = os.OpenFile(logf, os.O_RDWR|os.O_TRUNC|os.O_CREATE, 0644); err != nil {
+		log.Infof("Failed to open %s for write: %v", logf, err)
+		return nilCmd, err
+	}
+
+	// It would be nice to only enable debug (or maybe to only log to file at all)
+	// if 'stacker --debug', but we do not have access to that info here.
+	// to debug squashfuse, use "allow_other,debug"
+	cmd := exec.Command(sqfuse, "-f", "-o", "allow_other,debug", squashFile, extractDir)
+	cmd.Stdin = nil
+	cmd.Stdout = cmdOut
+	cmd.Stderr = cmdOut
+	cmdOut.Write([]byte(fmt.Sprintf("# %s\n", strings.Join(cmd.Args, " "))))
+	log.Debugf("Extracting %s -> %s with squashfuse [%s]", squashFile, extractDir, logf)
+	err = cmd.Start()
+	if err != nil {
+		return nilCmd, err
+	}
+
+	return cmd, nil
+}
+
 func ExtractSingleSquash(squashFile string, extractDir string, storageType string) error {
 	err := os.MkdirAll(extractDir, 0755)
 	if err != nil {
@@ -226,35 +272,11 @@ func ExtractSingleSquash(squashFile string, extractDir string, storageType strin
 		return err
 	}
 
-	findSqfusePath := func() string {
-		if p := which("squashfuse_ll"); p != "" {
-			return p
-		}
-		return which("squashfuse")
+	_, err = squashFuse(squashFile, extractDir)
+	if err == nil || err != squashNotFound {
+		return err
 	}
-
-	if sqfuse := findSqfusePath(); sqfuse != "" {
-		// given extractDir of path/to/some/dir[/], log to path/to/some/.dir-squashfs.log
-		extractDir := strings.TrimSuffix(extractDir, "/")
-
-		var cmdOut io.Writer
-		logf := path.Join(path.Dir(extractDir), "."+path.Base(extractDir)+"-squashfuse.log")
-		if cmdOut, err = os.OpenFile(logf, os.O_RDWR|os.O_TRUNC|os.O_CREATE, 0644); err != nil {
-			log.Infof("Failed to open %s for write: %v", logf, err)
-			return err
-		}
-
-		// It would be nice to only enable debug (or maybe to only log to file at all)
-		// if 'stacker --debug', but we do not have access to that info here.
-		// to debug squashfuse, use "allow_other,debug"
-		cmd := exec.Command(sqfuse, "-f", "-o", "allow_other,debug", squashFile, extractDir)
-		cmd.Stdin = nil
-		cmd.Stdout = cmdOut
-		cmd.Stderr = cmdOut
-		cmdOut.Write([]byte(fmt.Sprintf("# %s\n", strings.Join(cmd.Args, " "))))
-		log.Debugf("Extracting %s -> %s with squashfuse [%s]", squashFile, extractDir, logf)
-		return cmd.Start()
-	} else if p := which("unsquashfs"); p != "" {
+	if p := which("unsquashfs"); p != "" {
 		log.Debugf("Extracting %s -> %s with unsquashfs -f -d %s %s", extractDir, squashFile, extractDir, squashFile)
 		cmd := exec.Command("unsquashfs", "-f", "-d", extractDir, squashFile)
 		cmd.Stdout = os.Stdout
@@ -297,7 +319,7 @@ func whichSearch(name string, paths []string) string {
 	var search []string
 
 	if strings.ContainsRune(name, os.PathSeparator) {
-		if path.IsAbs(name) {
+		if filepath.IsAbs(name) {
 			search = []string{name}
 		} else {
 			search = []string{"./" + name}
@@ -305,7 +327,7 @@ func whichSearch(name string, paths []string) string {
 	} else {
 		search = []string{}
 		for _, p := range paths {
-			search = append(search, path.Join(p, name))
+			search = append(search, filepath.Join(p, name))
 		}
 	}
 
--- a/pkg/squashfs/verity.go
+++ b/pkg/squashfs/verity.go
@@ -71,6 +71,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 	"unsafe"
 
 	"github.com/anuvu/squashfs"
@@ -78,6 +79,7 @@ import (
 	"github.com/martinjungblut/go-cryptsetup"
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
+	"stackerbuild.io/stacker/pkg/log"
 	"stackerbuild.io/stacker/pkg/mount"
 )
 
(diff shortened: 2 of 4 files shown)


squashfs/squashfs.go
```diff
@@ -299,5 +304,16 @@ func GuestMount(squashFile string, mountpoint string, rootHash string) error {
 	if err != nil {
 		return fmt.Errorf("Failed releasing squashfuse process: %w", err)
 	}
+
+	for !isMountpoint(extractDir) {
+		fmt.Printf("%s is not yet mounted...\n", extractDir)
+		time.Sleep(time.Second)
```
smoser: thanks for this.. I thought I had dome something, but can't find it. 

1 second is an eternity here. this probably happens in the time frame of small milliseconds, a one second penalty for missing once seems way high to me.

As a simpler poll, could you do get a fileInfo before calling squashfuse, and then loop over a Stat and call to [os.SameFile](https://pkg.go.dev/os#SameFile) ?

that would seem lighter to me than parsing /proc/mounts repeatedly.

squashfs/squashfs.go
```diff
@@ -284,30 +295,24 @@ func GuestMount(squashFile string, mountpoint string, rootHash string) error {
...
 	}
 
-	for !isMountpoint(extractDir) {
-		fmt.Printf("%s is not yet mounted...\n", extractDir)
-		time.Sleep(time.Second)
+	for count := 0; !isMountpoint(mountpoint); count++ {
+		if count % 10 == 0 {
+			log.Debugf("%s is not yet mounted...\n", mountpoint)
+		}
+		time.Sleep(time.Duration(100 * time.Millisecond))
```
smoser, line 315: ok, good, you shortened it here.

smoser (changes requested):
some suggestions / questions.
i do think the uid_map reading is wrong, so i'm saying 'Request changes'

squashfs/verity.go
```diff
@@ -200,7 +200,37 @@ func verityName(p string) string {
...
+		return false
+	}
+	lines := string(bytes)
+	if len(lines) != 1 {
+		return false
+	}
+	words := strings.Fields(lines)
+	if len(words) != 3 || words[0] != "0" || words[1] != "0" {
+		return false
+	}
```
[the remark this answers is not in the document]
hallyn: Your example won't fool it bc it only checks the first line.  However, root on the host *can* fool it:

```
lxc-usernsexec -m b:0:0:4294967295 bash
root@jerom:/home/serge# cat /proc/self/uid_map
         0          0 4294967295
```

We could check /proc/self/ns/user vs /proc/1/ns/user but that also won't work in most container cases.

So I think we just say "don't do that" for now.

We might want to consider adding a flag to /proc/self/stat or /proc/self/status to indicate "not in the initial userns"

squashfs/squashfs.go
hallyn: Oh, I like that SameFile idea, thanks.

squashfs/squashfs.go
```diff
@@ -252,3 +253,51 @@ func whichSearch(name string, paths []string) string {
...
+	// handle it
+	if p := which("squashfuse"); p == "" {
+		return fmt.Errorf("squashfuse not found")
+	}
+
+	extractDir := strings.TrimSuffix(mountpoint, "/")
+
+	var cmdOut io.Writer
+	var err error
+	logf := path.Join(path.Dir(extractDir), "."+path.Base(extractDir)+"-squashfuse.log")
```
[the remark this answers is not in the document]
hallyn: True.  We could
1. only do the debug stuff if a --debug flag is given, or
2. put all the logs under mountpoint/meta/
or both.

squashfs/verity.go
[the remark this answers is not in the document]
smoser: https://go.dev/play/p/_S4VKcRLweQ

i think mine does "trick" it, it comes back with false, but should say true.

atomfs/molecule.go
```diff
@@ -94,8 +95,26 @@ func (m Molecule) overlayArgs(dest string) (string, error) {
 // device exists. so try to cooperate via this lock.
 var advisoryLockPath = path.Join(os.TempDir(), ".atomfs-lock")
 
-func (m Molecule) Mount(dest string) error {
+func makeLock(mountpoint string) (*os.File, error) {
 	lockfile, err := os.Create(advisoryLockPath)
```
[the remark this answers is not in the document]
hallyn: Do we want to create a github issue to discuss a proper fix for this?

Actually perhaps we should lock the oci/index.json?  But that'll fail (i think) if it's readonly - i.e. an iso.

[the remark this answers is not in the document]
hallyn: Sorry, the above is nonsense.  We don't need to lock the source.  We should simply create $(mountpoint)/lock, and lock it until ${mountpoint} is overmounted.

squashfs/squashfs.go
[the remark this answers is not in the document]
hallyn: Really I'm not sure of the best way to go about this, and it doesn't seem al that important.

squashfs/verity.go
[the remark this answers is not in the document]
hallyn: Notes:
1. My code was in fact wrong in how it was splitting the lines.  I've now fixed it in the PR.
2. https://go.dev/play/p/_K3a_WfgmmW   Should give an accurate test.

rchincha:
Slated for v0.40.2

hallyn:
> using fmt.Errorf directive; use errors.Errorf instead

That is wrong guidance.  I see that it's hardcoded in test/static-analysis.sh, but as I don't have my nitrokey with me I can't append a patch to this PR to remove it.

pkg/squashfs/verity.go
```diff
@@ -200,7 +202,116 @@ func verityName(p string) string {
...
+}
+
+func isMountpoint(dest string) bool {
+	mounted, err := mount.IsMountpoint(dest)
+	return err == nil && mounted
+}
+
+// Takes /proc/self/uid_map contents as one string
+// Returns true if this is a uidmap representing the whole host
+// uid range.
```
smoser, line 270: oh! I thought you were checking to see if `mappedUidZero == HostUidZero` . 

I assumed that operating as HostUidZero was enough.

Here is a script I wrote that shows it is enough, at least for ext2 and if root owns the block device (or the group owner is also mapped in).

```bash
#!/bin/sh

MP=""
LODEV=""
TEMP_D=""
RESTORE_LODEV=""
INSIDE="false"

fail() { msg "$@"; exit 1; }
msg() { echo "$@" 1>&2; }
cleanup() {
    msg "cleaning up"
    if [ -n "$MP" ]; then
        r umount "$MP"
    fi
    if [ -n "$LODEV" ]; then
        r losetup -d "$LODEV"
    fi
    if [ -n "$RESTORE_LODEV" ]; then
        r chown "$RESTORE_LODEV" "$LODEV"
    fi
    if [ -n "$TEMP_D" ]; then
        r rm -Rf "$TEMP_D"
    fi
}

r() {
    local p="$"
    [ "$INSIDE" = "true" ] && p="%"
    msg "$p" "$@"
    "$@"
}

inside() {
    local dev="$1" mp="$2"
    INSIDE=true
    msg "Inside: dev=\"$dev\" mp=\"$mp\""
    r id || fail
    r ls -l "$dev" &&
        r ls -ld "$mp" || fail "strange, ls failed"
    r mount "$dev" "$mp" || fail "mount failed"
    MP="$mp"
    r touch "$mp/file1"
    r ls "$mp/file1"
}

trap cleanup EXIT
if [ "$1" = "inside" ]; then
    shift
    inside "$@"
    exit
fi
TEMP_D=$(mktemp -d)

[ "$(id -u)" = "0" ] || fail "gotta be root"
echo "Outside:"
id || fail

r grep root /etc/subuid || fail
r grep root /etc/subgid || fail

img="${TEMP_D}/fs.img"
mp="${TEMP_D}/mp"
mkdir "$mp" || fail "failed mkdir mp"
r truncate --size=100M "$img" || fail "failed create image $img"

r mkfs.ext2 "$img" >/dev/null 2>&1 || fail "failed mkfs $img"

LODEV=$(r losetup --find --show "$img") || fail "failed losetup"
msg "created lodev $LODEV"

olduid=$(stat "--format=%U" "$LODEV") || fail "faild to get owner of $LODEV"
oldgid=$(stat "--format=%G" "$LODEV") || fail "faild to get owner group of $LODEV"
chown 0:0 "$LODEV" || fail "failed chown 0:0 $LODEV"
RESTORE_LODEV="$olduid:$oldgid"

r "$0" inside "$LODEV" "$mp" || fail "didn't work"
```

output looks like this:

```
$ sudo /tmp/try-me 
Outside:
uid=0(root) gid=0(root) groups=0(root),130(sbuild)
$ grep root /etc/subuid
root:0:1
$ grep root /etc/subgid
root:0:1
$ truncate --size=100M /tmp/tmp.6BqyRL5rEI/fs.img
$ losetup --find --show /tmp/tmp.6BqyRL5rEI/fs.img
created lodev /dev/loop14
$ /tmp/try-me inside /dev/loop14 /tmp/tmp.6BqyRL5rEI/mp
Inside: dev="/dev/loop14" mp="/tmp/tmp.6BqyRL5rEI/mp"
% id
uid=0(root) gid=0(root) groups=0(root),130(sbuild)
% ls -l /dev/loop14
brw-rw---- 1 root root 7, 14 Dec 19 22:24 /dev/loop14
% ls -ld /tmp/tmp.6BqyRL5rEI/mp
drwxr-xr-x 2 root root 4096 Dec 19 22:24 /tmp/tmp.6BqyRL5rEI/mp
% mount /dev/loop14 /tmp/tmp.6BqyRL5rEI/mp
% touch /tmp/tmp.6BqyRL5rEI/mp/file1
% ls /tmp/tmp.6BqyRL5rEI/mp/file1
/tmp/tmp.6BqyRL5rEI/mp/file1
cleaning up
% umount /tmp/tmp.6BqyRL5rEI/mp
cleaning up
$ losetup -d /dev/loop14
$ chown root:disk /dev/loop14
$ rm -Rf /tmp/tmp.6BqyRL5rEI
```

pkg/squashfs/verity_test.go
```diff
@@ -11,6 +11,47 @@ import (
...
+	uidmapTestcase{
+		uidmap:   `         10          0 4294967295`,
+		expected: false,
+	},
+	uidmapTestcase{
+		uidmap:   `         0          10 4294967295`,
+		expected: false,
+	},
+	uidmapTestcase{
+		uidmap:   `         0          0 1`,
```
smoser, line 42: this is the case that I tested above (I think) and the result should be true.

pkg/squashfs/verity.go
hallyn: I'm missing something - where are you doing a userns unshare and mapping?  I see no lxc-usernsexec or unshare call in there.

If I edit your script to do:
```
-r "$0" inside "$LODEV" "$mp" || fail "didn't work"
+r unshare -U -r "$0" inside "$LODEV" "$mp" || fail "didn't work"
```
Then it fails with
```
% mount /dev/loop32 /tmp/tmp.oZx0FfP99s/mp
mount: /tmp/tmp.oZx0FfP99s/mp: permission denied.
mount failed
```
Now, it *is* interesting that losetup succeeded.  I didn't expect that.

smoser: bah, you're right. the last line of the script was supposed to say:
```r lxc-usernsexec -m b:0:0:1 -- "$0" inside "$LODEV" "$mp" || fail "didn't work"```

so yeah, that doesn't work. But why not?

Comment (hallyn) [reply pkg/squashfs/verity.go:270]: Because to have capable(CAP_SYS_ADMIN), which equals ns_capable(init_user_ns, CAP_SYS_ADMIN), you (a) must have CAP_SYS_ADMIN in your pE (which you do), and (b) must be in a user_ns which is same as or ancestor to the target user_ns, init_user_ns.  Your userns is a child of it.  You cannot obtain capable(CAP_ANYTHING).

**Review by smoser — COMMENTED** [review]
There are some issues that need fixing.

Would it be easier / better to have 'guestMount' return a known error? 

```golang
noSquashFuseAvailable = errors.New("No squashfuse binary in PATH")
```
And then callers could handle that error specially ?

pkg/squashfs/verity_test.go
hallyn: iiuc we're now agreed that this is not the case.

pkg/squashfs/verity.go
```diff
@@ -200,7 +202,116 @@ func verityName(p string) string {
...
+	}
+	fiPre, err := os.Lstat(mountpoint)
+	if err != nil {
+		return fmt.Errorf("Failed stat'ing %q: %w", mountpoint, err)
+	}
+	if fiPre.Mode()&os.ModeSymlink != 0 {
+		return fmt.Errorf("Refusing to mount onto a symbolic linkd")
+	}
+
+	cmd, err := squashFuse("", squashFile, mountpoint)
```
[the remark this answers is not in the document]
hallyn: Oops.  Yeah - merging my change with your latest change to your squashfs.go patch didn't go right, I guess.  Originally I had squashFuse finding the path itself and the callers just checked whether it existed.

I still think I should really have squashFuse be called unconditionally without passing in the cmdPath, and have squashFuse return a custom-made squashfuse-not-found error if not found.  So maybe I'll change it to that.

pkg/squashfs/squashfs.go
```diff
@@ -214,6 +215,44 @@ func maybeKernelSquashMount(squashFile, extractDir string) (bool, error) {
...
 }
 
+func findSquashfusePath() string {
+	if p := which("squashfuse_ll"); p != "" {
+		return p
+	}
+	return which("squashfuse")
+}
+
+func squashFuse(cmdPath, squashFile, extractDir string) (*exec.Cmd, error) {
```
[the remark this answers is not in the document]
hallyn: I agree.

hallyn:
> There are some issues that need fixing.
> 
> Would it be easier / better to have 'guestMount' return a known error?
> 
> ```go
> noSquashFuseAvailable = errors.New("No squashfuse binary in PATH")
> ```
> 
> And then callers could handle that error specially ?

Oh - that's what I ended up doing.

pkg/squashfs/verity.go
```diff
@@ -200,7 +202,114 @@ func verityName(p string) string {
...
+	}
+	squashFile = abs
+
+	abs, err = filepath.Abs(mountpoint)
+	if err != nil {
+		return errors.Errorf("Failed to get absolute path for %s: %v", mountpoint, err)
+	}
+	mountpoint = abs
+
+	cmdPath := findSquashfusePath()
```
smoser, line 232: i think this is supposed to change to inspecting the returned error.

hallyn: Yup, thanks.

PR #361, opened 2022-12-13
Use kernel mount of squashfs if running privledged.

smoser:
If stacker build is being run as "real root", then use a kernel mount rather than using squashfuse.

**What does this PR do / Why do we need it**: Maybe we dont.  It seems unreasonable to hamstring root into only doing fuse mounts when he/she could have the kernel do them.

I expect some discussion would be needed here.  I'm thinking that maybe environment variable configuration to turn this on or off would be useful.  Maybe default to fuse and allow the root user of stacker to use kernel mounts by setting STACKER_SQUASHFS_KERNEL_MOUNTS=true . or something.

### DIFF
--- a/squashfs/squashfs.go
+++ b/squashfs/squashfs.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"strings"
 	"sync"
+	"syscall"
 
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
@@ -19,6 +20,9 @@ import (
 
 var checkZstdSupported sync.Once
 var zstdIsSuspported bool
+
+var tryKernelMountSquash bool = true
+var kernelSquashMountFailed error = errors.New("kernel squash mount failed")
 
 // ExcludePaths represents a list of paths to exclude in a squashfs listing.
 // Users should do something like filepath.Walk() over the whole filesystem,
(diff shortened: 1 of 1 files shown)


squashfs/squashfs.go
```diff
@@ -175,7 +175,14 @@ func ExtractSingleSquash(squashFile string, extractDir string, storageType strin
 		return which("squashfuse")
 	}
 
-	if sqfuse := findSqfusePath(); sqfuse != "" {
+	if realuid := os.Getenv("STACKER_REAL_UID"); realuid == "0" {
```
hallyn, line 178: If I'm running stacker inside an unprivileged container as root, my STACKER_REAL_UID will be 0 right?

What if we ... just try it?  Try the mount -o loop, if cmd.Run() fails with EPERM, then (a) log that and don't try again, and (b) try squashfuse.

squashfs/squashfs.go
```diff
@@ -162,12 +166,66 @@ func MakeSquashfs(tempdir string, rootfs string, eps *ExcludePaths, verity Verit
...
+	return false, kernelSquashMountFailed
+}
+
 func ExtractSingleSquash(squashFile string, extractDir string, storageType string) error {
 	err := os.MkdirAll(extractDir, 0755)
 	if err != nil {
 		return err
 	}
 
+	if mounted, err := maybeKernelSquashMount(squashFile, extractDir); err != nil && mounted {
```
hallyn, line 223: did you really mean err != nil here?

hallyn (approved):

PR #362, opened 2022-12-13
build: also build libdevmapper in parallel

rchincha:
PR #258 resurrected

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -25,6 +25,8 @@ build-env:
       squashfs-tools-ng-dev \
       linux-headers
 
+    numcpu=$(grep -c processor /proc/cpuinfo)
+
     # json-c doesn't have static binaries in alpine
     apk add cmake
     tar -xf /stacker/json-c-*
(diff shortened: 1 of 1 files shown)


build.yaml
```diff
@@ -39,7 +39,7 @@ build-env:
     tar -xf /stacker/v2_03_15.tar.gz
     cd lvm2-*
     ./configure --enable-static_link
-    make install_device-mapper
+    make -j$(grep -c processor /proc/cpuinfo) install_device-mapper
```
smoser, line 42: fwiw, this just silently-ish swallows failure of the 'grep' command.

better would be:

```sh
numcpu=$(grep -c processor /proc/cpuinfo)
make -j${numcpu} install_device-mapper
```

example:

```
$ cat /tmp/foo
#!/bin/sh
set -e
echo make -j$(bogus command) foo
echo still here

$ sh -x /tmp/foo
+ set -e
+ bogus command
/tmp/foo: 3: bogus: not found
+ echo make -j foo
make -j foo
+ echo still here
still here
```

you end up calling make like `make -j install_device-mapper`.  make will probably fail, but its just odd.

smoser (changes requested):

build.yaml
rchincha: fixed

rchincha:
/usr/lib/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/lib/gcc/x86_64-alpine-linux-musl/12.2.1/../../../../x86_64-alpine-linux-musl/bin/ld: cannot find -ldevmapper: No such file or directory
collect2: error: ld ret


Not sure devmapper code can handle parallel make

rchincha:
Closing this PR for now.

Issue #377, opened 2022-12-22
Feat: just specify a single `--dir` under which other working dirs can be created


rchincha:
### Is your feature request related to a problem? Please describe.

Currently, a roots-dir, oci-dir and stacker-dir need to be specified. While we can still keep those, more convenient to just specify one dir.

PR #378, opened 2022-12-23
feat: add a single working-dir option

rchincha:
Fixes issue #377

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/cmd/stacker/main.go
+++ b/cmd/stacker/main.go
@@ -122,6 +122,10 @@ func main() {
 
 	app.Flags = []cli.Flag{
 		cli.StringFlag{
+			Name:  "work-dir",
+			Usage: "set the working directory for stacker's cache, OCI output and rootfs output",
+		},
+		cli.StringFlag{
 			Name:  "stacker-dir",
 			Usage: "set the directory for stacker's cache",
 			Value: ".stacker",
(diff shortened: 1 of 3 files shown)


smoser (commented):
Does this work?
Nothing changed seems to reference the newly-added WorkingDir, so I don't know how it would.

hallyn:
> 

Yeah that's odd - was a hunk of patch (where the stackerdir etc are set relative to workingdir if the former is not set and the latter is) accidentally dropped?

cmd/stacker/main.go
```diff
@@ -225,13 +229,30 @@ func main() {
 
 		config.EmbeddedFS = embeddedFS
 
-		if config.StackerDir == "" || ctx.IsSet("stacker-dir") {
+		if ctx.IsSet("work-dir") {
+			config.WorkDir = ctx.String("work-dir")
+		}
+		if ctx.IsSet("stacker-dir") {
+			config.StackerDir = ctx.String("stacker-dir")
+		} else if ctx.IsSet("work-dir") {
```
hallyn, line 237: These three hunks don't make sense to me.  Does 'ctx.IsSet(X)' mean something other than what I'd think?

You're saying
```
if ctx.IsSet("stacker-dir) {
   config.StackerDir = ctx.String("stacker-dir")
} else if ctx.IsSet("work-dir") {
  config.StackerDir = path.Join(ctx.String("work-dir"), ctx.String("stacker-dir")) // but ctx.IsSet("stacker-dir") is false!
```

rchincha: https://github.com/project-stacker/stacker/pull/378#issuecomment-1407024467

^ trying to achieve this

hallyn: Oh, I see - IsSet() threw me off because it's smarter than I thought.  ctx.String("stacker-dir") will return the default value while ctx.IsSet("stacker-dir") returns "false.

Additionally, 'work-dir' left me confused last night as it makes me think of the overlay 'work-dir'.  But I don't have a better suggestion.

smoser (approved):
lgtm

Issue #370, opened 2022-12-16
Bug: top level directories are removed if there is only one


smoser:
### stacker version

v0.30.1-15-gdc0ada0

### Describe the bug

If you attempt to build the stacker.yaml below it will fail.
The reason is that the `/files` directory is stripped away in the 'mybuild'  layer.  The files 'subuid' and 'subgid' are directly in /.

    $ stacker build --layer-type=squashfs --layer-type=tar
    ...
    error: lstat /files/subuid: no such file or directory
    error: execute failed: exit status 1
    error: exit status 1

```yaml
$ cat stacker.yaml 
mybuild:
  from:
    type: scratch
  import:
    - path: /etc/subuid
      dest: /files/
    - path: /etc/subgid
      dest: /files

mytest:
  from: 
    type: docker
    url: docker://ubuntu:latest
  import:
    - stacker://mybuild/files/subuid
    - stacker://mybuild/files/subgid
  run: |
    ls -l /stacker
```

smoser:
Part of the fix is this:
```
--- a/pkg/squashfs/squashfs.go
+++ b/pkg/squashfs/squashfs.go
@@ -135,7 +135,7 @@ func MakeSquashfs(tempdir string, rootfs string, eps *ExcludePaths, verity Verit
        tmpSquashfs.Close()
        os.Remove(tmpSquashfs.Name())
        defer os.Remove(tmpSquashfs.Name())
-       args := []string{rootfs, tmpSquashfs.Name()}
+       args := []string{rootfs, tmpSquashfs.Name(), "-keep-as-directory"}
        compression := GzipCompression
        if mksquashfsSupportsZstd() {
                args = append(args, "-comp", "zstd")
```

but I think the problem is also happening with tar layers. (fails to build with '--layer-type=tar' also).

smoser:
When I was poking around trying to fix this, I noticed that each 'import' gets creates its own layer in the created oci. 
i think it might be better to have stacker start populating the same overlay that the 'run' would be handed before 'run' (if present) occurs.

Given this stcker.lyaml:
```yaml
mybuild:
  from:
    type: docker
    url: ubuntu:latest
  import:
    - my-file
    - path: /etc/passwd
      dest: /etc/passwd-host
    - path: /static/some-binary-here
      dest: /usr/bin/
  run: |
    ls -l /usr/bin/some-binary-here
    ls -l /etc/passwd /etc/passwd-host
    ls -l /stacker/my-file
```

[...]

PR #387, opened 2023-01-09
fix: don't use quoting for paths during import

rchincha:
Fixes issue #370

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/test/import.bats
+++ b/test/import.bats
@@ -345,6 +345,23 @@ fourth:
         uid: 1000
         gid: 1000
         dest: /usr/bin
+fifth:
+  from:
+    type: scratch
+  import:
+    - path: test_file
+      dest: /files/
+    - path: test_file2
+      dest: /files
+sixth:
+  from:
+    type: docker
+    url: docker://ubuntu:latest
+  import:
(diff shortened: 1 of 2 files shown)


rchincha:
@smoser your exact same test case in issue #370 has been included in this PR

test/import.bats
```diff
@@ -345,6 +345,23 @@ fourth:
...
         gid: 1000
         dest: /usr/bin
+fifth:
+  from:
+    type: scratch
+  import:
+    - path: /etc/subuid
+      dest: /files/
+    - path: /etc/subgid
+      dest: /files
```
smoser, line 355: you should probably use paths that are created in the test case. I picked /etc/subuid and /etc/subgid because they were kind of known to be there, but for the test better to just create some file paths and use those like is done elsewhere.

smoser (changes requested):
:tada: Nice work @rchincha .

I do have one request for a change, but good job.

test/import.bats
rchincha: fixed

smoser (approved):
thank you.

Issue #388, opened 2023-01-10
Bug: overlay,userxattr directory created in roots/<layer>/ during squashfs biulds


smoser:
### stacker version

v0.40.1-e54a685

### Describe the bug

```
$ stacker build
preparing image minbase...
loading docker://alpine:edge
Copying blob a71ff8a5264a done  
Copying config 09b6b3579b done  
Writing manifest to image destination
Storing signatures
+ echo all done.
all done.

$ ls -l roots/minbase/
total 24
drwxr-xr-x 3 smoser smoser 4096 Jan 10 09:07 overlay
-rw-r--r-- 1 smoser smoser  458 Jan 10 09:07 overlay_metadata.json
drwxr-xr-x 2 smoser smoser 4096 Jan 10 09:07 overlay,userxattr
drwxr-xr-x 2 smoser smoser 4096 Jan 10 09:07 rootfs
drwxr-xr-x 3 smoser smoser 4096 Jan 10 09:07 work
drwxr-xr-x 2 smoser smoser 4096 Jan 10 09:07 workaround0
```

The 'overlay' dir is probably not necessary to be there (stacker could just create a tmpdir always on mount), but the 'overlay,usrexattr' dir is I think fallout of `GetLXCRootfsConfig` in `pkg/overlay/overlay.go` doing :

```golang
    return fmt.Sprintf("overlay:%s,userxattr", lxcRootfsString), nil 
```

I think that lxc is not understanding the intent of userxattr there.

See stacker.yaml below.

```yaml
$ cat stacker.yaml 
minbase:
  build_only: true
  from:
    type: docker
    url: ${{DOCKER_BASE:docker://}}alpine:edge
  run: |
    echo all done.
```

smoser:
I think #389 is intended to fix this.

hallyn:
Assigning this to myself, but not sure when I'll have a chance to look, so anyone please feel free to assign yourself if you do happen to start on it.

smoser:
is it just a matter of removing the ,userxattr ?

hallyn:
Oh we're doing that for lxc.rootfs string?  Then yes, I think so.

PR #389, opened 2023-01-10
fix: update lxd dependency

hallyn:
Update lxd to one including:

commit 599dd2280553f816b945ebf5a128a6fc690a5219
Author: Andreas Dilger <adilger@whamcloud.com>
Date:   Fri Jan 7 12:50:05 2022 -0700

    shared/util/linux: only complain on xattr size increase

As otherwise reading xattrs on overlay can break.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,8 +12,8 @@ require (
 	github.com/freddierice/go-losetup v0.0.0-20220331141030-7ad00c49b973
 	github.com/justincormack/go-memfd v0.0.0-20170219213707-6e4af0518993
 	github.com/klauspost/pgzip v1.2.5
-	github.com/lxc/go-lxc v0.0.0-20210607135324-10de240d43ab
-	github.com/lxc/lxd v0.0.0-20211118162824-0a8d8c489961
+	github.com/lxc/go-lxc v0.0.0-20220627182551-ad3d9f7cb822
+	github.com/lxc/lxd v0.0.0-20230109185737-f7ccf0330640
 	github.com/martinjungblut/go-cryptsetup v0.0.0-20220314071908-7b9938e4a08c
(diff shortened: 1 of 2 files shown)


smoser (changes requested):
I think you're intending that this 'Fixes #388', right?

You should put that in the commit message.

hallyn:
> I think you're intending that this 'Fixes #388', right?
> 
> You should put that in the commit message.

Hm, no, I don't think that's related.

This is the fix for overlay's listxattr(path, NULL, 0) returning a larger size estimate than the actual size returned when you pass a buffer with listxattr(path, buf, sz).  When you pass a non-zero sz (and a buffer), then overlay filters out some xattrs.  When only telling you the size, it doesn't filter those out, so the size you're told to allocate can be more than what you really need.  We shouldn't fail when that happens.

smoser:
> > I think you're intending that this 'Fixes #388', right?
> > You should put that in the commit message.
> 
> Hm, no, I don't think that's related.
> 
> This is the fix for overlay's listxattr(path, NULL, 0) returning a larger size estimate than the actual size returned when you pass a buffer with listxattr(path, buf, sz). When you pass a non-zero sz (and a buffer), then overlay filters out some xattrs. When only telling you the size, it doesn't filter those out, so the size you're told to allocate can be more than what you really need. We shouldn't fail when that happens.

oh, sorry then. i guess the timing of this after my open of #388 made me think that newer lxc would dtrt with that ',userxattr'. and the commit message even inculded 'xattr'.  just bad luck. sorry for the noise.

hallyn:
Ok wel lthis is also folded into #391 so i'll close this

PR #391, opened 2023-01-10
Support optional dmverity when remounting as well

hallyn:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


mikemccracken (approved):

hallyn:
Ok, now we're at a point where we have a decision to make.

cmd/stacker/internal_go.go:doAtomfsMount() sets AllowMissingVerityData: true.

Then test/atomfs.bats:141 does
bad_stacker internal-go atomfs mount test-squashfs mountpoint | grep "invalid root hash"

Since AllowMissingVerityData is true, my patch is making it ignore missing roothash.  The reason my patch needs to do that is because otherwise when trying to verify the roothash, the ioctl fails on squashfuse.

Ok so I guess I need to tweak the internal api a bit more so that only failure of the ioctl is allowed with AllowMissingVerityData==true, but if we *can* get roothash from the mounted device, then even if AllowMissingVerityData==true, we should fail.

I'll put this in draft mode until I can fix this, which should be tomorrow morning.

PR #411, opened 2023-02-13
feat(convert): export ENV vars

rchincha:
We could have passed these into the "build_env:" directive, however it appears environment substitution is also allowed such as PATH=/usr/local/sbin:$PATH for that layer, which is harder to figure out. Instead just "export" them in the "run:" directive.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/stacker/convert.go
+++ b/pkg/stacker/convert.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -32,6 +33,7 @@ type Converter struct {
 	currLayer string
 	subs      map[string]string
 	args      []string
+	env       map[string]string
 	// per-layer state
 	currDir string
 	currUid string
@@ -98,6 +100,11 @@ func (c *Converter) convertCommand(cmd *Command) error {
 		}
 		c.output[c.currLayer] = &layer
 	case "run":
+		// setup the environment
+		for k, v := range c.env {
+			layer.Run = append(layer.Run, fmt.Sprintf("export %s=%s", k, v))
+		}
+
 		// replace any ARGs first
 		for i, val := range cmd.Value {
 			for _, arg := range c.args {
@@ -123,7 +130,8 @@ func (c *Converter) convertCommand(cmd *Command) error {
 
 		for _, line := range cmd.Value {
 			// patch some cmds
-			line = strings.ReplaceAll(line, "mkdir", "mkdir -p ")
+			re := regexp.MustCompile(`\bmkdir\b`)
+			line = re.ReplaceAllString(line, "mkdir -p")
 
 			if c.currUid == "" {
 				// picking 'bash' here
@@ -183,11 +191,11 @@ func (c *Converter) convertCommand(cmd *Command) error {
 			}
 		}
 
-		if layer.Environment == nil {
-			layer.Environment = map[string]string{}
+		if c.env == nil {
+			c.env = map[string]string{}
 		}
 
-		layer.Environment[cmd.Value[0]] = val
+		c.env[cmd.Value[0]] = val
 	case "workdir":
 		layer.Run = append(layer.Run, fmt.Sprintf("cd %s", cmd.Value[0]))
 		c.currDir = cmd.Value[0]
--- a/test/convert.bats
+++ b/test/convert.bats
@@ -30,37 +30,11 @@ EOF
   # build should now work
   ## docker build -t test
   mkdir -p /out
-  stacker build -f stacker.yaml --substitute-file stacker-subs.yaml
-  rm -f stacker.yaml stacker-subs.yaml
-  stacker clean
-}
-
-@test "build alpine image" {
-  git clone https://github.com/alpinelinux/docker-alpine.git
-  chmod -R a+rwx docker-alpine
-  cd docker-alpine
-  mkdir -p /out
-  chmod -R a+rwx /out
-  stacker convert --docker-file Dockerfile --output-file stacker.yaml --substitute-file stacker-subs.yaml
-  stacker build -f stacker.yaml --substitute-file stacker-subs.yaml --substitute IMAGE=alpine
+  stacker build -f stacker.yaml --substitute-file stacker-subs.yaml --substitute IMAGE=app
   if [ -z "${REGISTRY_URL}" ]; then
-      skip "skipping test because no registry found in REGISTRY_URL env variable"
+    skip "skipping test because no registry found in REGISTRY_URL env variable"
   fi
-  stacker publish -f stacker.yaml --substitute-file stacker-subs.yaml --substitute IMAGE=alpine --skip-tls --url docker://${REGISTRY_URL} --tag latest
-  rm -f stacker.yaml stacker-subs.yaml
-  stacker clean
-}
-
-@test "build elasticsearch image" {
-  git clone https://github.com/elastic/dockerfiles.git
-  chmod -R a+rwx dockerfiles
-  cd dockerfiles/elasticsearch
-  stacker convert --docker-file Dockerfile --output-file stacker.yaml --substitute-file stacker-subs.yaml
-  stacker build -f stacker.yaml --substitute-file stacker-subs.yaml --substitute IMAGE=elasticsearch
-  if [ -z "${REGISTRY_URL}" ]; then
-      skip "skipping test because no registry found in REGISTRY_URL env variable"
-  fi
-  stacker publish -f stacker.yaml --substitute-file stacker-subs.yaml --substitute IMAGE=elasticsearch --skip-tls --url docker://${REGISTRY_URL} --tag latest
+  stacker publish -f stacker.yaml --substitute-file stacker-subs.yaml --substitute IMAGE=app --skip-tls --url docker://${REGISTRY_URL} --layer app --tag latest
   rm -f stacker.yaml stacker-subs.yaml
   stacker clean
 }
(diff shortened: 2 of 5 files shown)


test/convert.bats
```diff
@@ -64,3 +64,17 @@ EOF
   rm -f stacker.yaml stacker-subs.yaml
   stacker clean
 }
+
+@test "build python image" {
+  git clone https://github.com/docker-library/python.git
+  chmod -R a+rwx python
```
smoser, line 70: this seems questionable.
you really want to make *everything* under there executable?

test/convert.bats
```diff
@@ -64,3 +64,17 @@ EOF
   rm -f stacker.yaml stacker-subs.yaml
   stacker clean
 }
+
+@test "build python image" {
+  git clone https://github.com/docker-library/python.git
```
smoser, line 69: Some day you'll wake up and this repository will have changed and tests will fail.  Is that desirable?
The alternative is to put the test you need here in the repository, which will be more reliable and faster, no?

test/convert.bats
```diff
@@ -64,3 +64,17 @@ EOF
...
+  chmod -R a+rwx python
+  cd python/3.11/alpine3.17
+  stacker convert --docker-file Dockerfile --output-file stacker.yaml --substitute-file stacker-subs.yaml
+  stacker build -f stacker.yaml --substitute-file stacker-subs.yaml --substitute IMAGE=python
+  if [ -z "${REGISTRY_URL}" ]; then
+      skip "skipping test because no registry found in REGISTRY_URL env variable"
+  fi
+  stacker publish -f stacker.yaml --substitute-file stacker-subs.yaml --substitute IMAGE=python --skip-tls --url docker://${REGISTRY_URL} --tag latest
```
smoser, line 77: My fault for not paying attention, but it seems like 'REGISTRY_URL' should be a URL, not a hostname/path (I'm assuming that a valid setting would be 'docker.io/library', which isn't a url at all).

smoser (commented):
This is an interesting change.
If I understand it correctly, it is so that you can support the following line from [here](https://github.com/docker-library/python/blob/master/3.11/alpine3.17/Dockerfile#L10):

    ENV PATH /usr/local/bin:$PATH

What is most surprising about it is that it seems to me that the above line is almost certainly *wrong*.  If it is setting the PATH environment to be used inside the container, then why would you consult the *hosts* value of PATH?  It should just explicitly set the value for the container like:

    ENV PATH usr/local/bin:/usr/sbin:/usr/bin

If my host had a PATH variable of `/bin` or perhaps '/opt/bin:/opt/sbin', that could be perfectly fine for my host.  But passing that variable through at best makes the execution less declarative and at worst breaks the build.

Maybe I'm missing something?

rchincha:
> This is an interesting change. If I understand it correctly, it is so that you can support the following line from [here](https://github.com/docker-library/python/blob/master/3.11/alpine3.17/Dockerfile#L10):
> 
> ```
> ENV PATH /usr/local/bin:$PATH
> ```
> 
> What is most surprising about it is that it seems to me that the above line is almost certainly _wrong_. If it is setting the PATH environment to be used inside the container, then why would you consult the _hosts_ value of PATH? It should just explicitly set the value for the container like:
> 
> ```
> ENV PATH usr/local/bin:/usr/sbin:/usr/bin
> ```
> 
> If my host had a PATH variable of `/bin` or perhaps '/opt/bin:/opt/sbin', that could be perfectly fine for my host. But passing that variable through at best makes the execution less declarative and at worst breaks the build.
> 
> Maybe I'm missing something?

The $PATH is derived from the FROM section layer. So how does one figure that out?

test/convert.bats
rchincha: This was to make sure the **exact** upstream tests work.

test/convert.bats
rchincha: This was borrowed from test/publish.bats

test/convert.bats
rchamarthy: Is it better to fork and then use the fork?

test/convert.bats
rchincha: .github/workflows/build.yaml
49:          REGISTRY_URL: localhost:5000

test/convert.bats
rchincha: fixed

rchincha: mkdir /home/runner/work/stacker/stacker/stackertest-test_build_python_image.2Okow7/python/3.11/alpine3.17/roots: permission denied couldn't make rootfs dir for testsuite

^ but this. Fixing.

test/convert.bats
smoser: yeah, we should fix that to be an actual URL, then it would or could support REGISTRY_URL=oci:mirrors/docker.io/library .  I realize you didn't change it here.

test/convert.bats
smoser: I can understand the desire to make sure that "real world examples" work.  I personally hate it when the internet breaks my c-i.  It means someone has to spend time figuring out why the light went from green to red, or worse, people accept "sometimes c-i breaks".

If you want to have a collection of "moving target real world docker files", maybe that is best to do in an integration test of some sort.  Unfortunately that just kicks the "figure out what failed" problem down the line.

It is entirely possible that a stacker change regresses this behavior and this test catches that.  But I think it is more likely that a breakage occurs due to the added dependency on a moving third party.

I would much prefer a quicker that did something like:

```
@test "check environment variable substitution" {
   cat > dockerfile <<EOF
FROM alpine:3.17

ENV PATH /usr/local/bin:$PATH

RUN [ "$PATH" = "/usr/local/bin:/usr/bin:/usr/sbin" ]
EOF
}
```

I like the REGISTRY_URL so that these tests don't have to depend on docker.io.

test/convert.bats
smoser: is that an indication of a larger problem with permissions? It seems unlikely that the problem here is with that specific directory and not more generic.

stacker should only need write permission for the uid owner of that directory in order to create 'roots' (and .stacker and oci) in the directory. 

what was the ownership of alpine3.17/ ?  How does content for the test get copied into that temp dir? maybe umask got foobar'ed somehow?

rchincha: Let's find out.

test/convert.bats
rchincha: That Dockerfile in that repo unlikely to change for a long while and if it breaks or more likely when it disappears, we will look at it and fix.

rchincha:
> That makes me feel much better, that it isn't just taking the current environment's value.

Yup.

test/convert.bats
smoser: > That Dockerfile in that repo unlikely to change for a long while and if it breaks or more likely when it disappears, we will look at it and fix.

I disagree with that statement.  It has [changed 5 times in 2 weeks](https://github.com/docker-library/python/commits/0a56f04a07495e0ed61663c73bc0c2e175773ddc/3.11/alpine3.17/Dockerfile) including a revert.

I really think it is very high cost in build time (we're building *python* as part of stacker c-i?) and reliability.  Successful c-i now depends on:
 * access to 'alpine:3.17' that will change over time
 * 'git clone' from github (which does not auto-retry on failure) and basically cannot be mirrored

And all of this is to check the expansion of a templated string.

test/convert.bats
rchincha:
```
# Cloning into 'python'...
# python/:
# total 104
# drwxr-xr-x 10 root   root   4096 Feb 14 17:37 .
# drwx------  5 runner runner 4096 Feb 14 17:37 ..
# drwxr-xr-x  8 root   root   4096 Feb 14 17:37 .git
# -rw-r--r--  1 root   root     94 Feb 14 17:37 .gitattributes
# drwxr-xr-x  3 root   root   4096 Feb 14 17:37 .github
# -rw-r--r--  1 root   root     17 Feb 14 17:37 .gitignore
```

Everything is root-owned, that's why

rchincha:
Let's split this into two parts. A nightly "convert/build" job that works on real-world examples, while we do basic tests under bats.

test/convert.bats
rchincha: @smoser, moved the conversion out of the core ci path, and add a "nightly" conversion workflow, at least some place we can catch issues.

pkg/stacker/convert.go
```diff
@@ -123,7 +129,7 @@ func (c *Converter) convertCommand(cmd *Command) error {
 
 		for _, line := range cmd.Value {
 			// patch some cmds
-			line = strings.ReplaceAll(line, "mkdir", "mkdir -p ")
+			line = strings.ReplaceAll(line, "mkdir", "mkdir -p")
```
hallyn, line 132: So if the dockerfile was doing

```
alias mkdir_p="mkdir -p $*"
mkdir_p /tmp/a
```
This will trip up.  Do we care?

Why is this needed - does a dockerfile actually do this?  If so, maybe doing a regexp replace for "\<mkdir\>" would at least be safer.

rchincha: Yes, we do need it to be compatible with docker behavior. But you also have a point.

rchincha: Simplest change maybe to 's/mkdir /mkdir -p /g'

hallyn: I was thinking something like

```
re := regexp.MustCompile('\\bmkdir\\b')
line  = re.ReplaceAllString(line, 'mkdir -p')

rchincha: Ok, your regexp suggestion/change went through

hallyn (approved):

pkg/stacker/convert.go
hallyn: cool, thanks.

PR #414, opened 2023-02-14
test: add more UTs

rchincha:

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/lib/file_test.go
+++ b/pkg/lib/file_test.go
@@ -1,0 +1,53 @@
+package lib_test
+
+import (
+	"io/fs"
+	"os"
+	"path"
+	"testing"
+
+	. "github.com/smartystreets/goconvey/convey"
+	"stackerbuild.io/stacker/pkg/lib"
+)
+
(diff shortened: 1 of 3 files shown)


pkg/lib/file_test.go
```diff
@@ -0,0 +1,47 @@
...
+			defer os.Remove(dest.Name())
+
+			mode := fs.FileMode(0644)
+			err = lib.FileCopy(dest.Name(), src.Name(), &mode, -1, -1)
+			So(err, ShouldBeNil)
+		})
+	})
+
+	Convey("FindFiles", t, func() {
+		files, err := lib.FindFiles("/tmp", ".*")
```
smoser, line 43: Can you create directory with os.MkdirTemp and then put a file in it and then use lib.FindFiles?  that way it won't depend on the assumption that /tmp exists on the system.

rchincha: fixed.

hallyn (approved):

PR #417, opened 2023-02-22
feat(run): allow choosing a different shell for `run:`

rchincha:
Currently, we hardcode to a default shell which may not work for all container images. So add a cmdline arg to override the default.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


hallyn:
I think allowing the person running stacker to specify the shell is a bad idea.  They're unlikely to know all the little bash- or zsh-specific things which are there.  You can already specify the shell you want to use by doing

```
  run: |
    #!/bin/bash -e
    echo hello world
```

What's the use case that this would fix which that does not?

Issue #311, opened 2022-09-27
add a "--verify" option for stacker builds


rchincha:
Add a `stacker build --verify` so that when the build encounters a `from:` remote url, the build will proceed only if from signed/trusted sources. The following options must be supported.

References:

`docker trust inspect alpine:edge`
https://github.com/docker/cli/blob/master/cli/command/trust/inspect.go#L22

https://github.com/sigstore/cosign

https://github.com/notaryproject/notation

rchincha:
stacker already supports a `insecure:` directive

insecure should mean:
1. try https best effort - bad/unknown CA ignored
2. else fallback to http
3. also ignore signatures

rchincha:
Also, `insecure: false` by default

smoser:
> stacker already supports a `insecure:` directive
> 
> insecure should mean:
> 
>     1. try https best effort - bad/unknown CA ignored
>     2. else fallback to http
>     3. also ignore signatures

That is a lot of things for one flag.
perhaps something more fine-grained?

I would also like to request that we avoid negative terms.  Ie, i'd rather have the keyword 'secure' and it default to 'true' than 'insecure' and it default to 'false'. doulbe negatives just cause confusion.

PR #419, opened 2023-02-27
feat(insecure): ensure builds are secure by default

rchincha:
BREAKING CHANGE: `insecure:` directive is now `true` by default. Previously, it was `false`.

Fixes issue #311

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


rchincha:
https://sse-secure-systems.github.io/connaisseur/v2.0.0/validators/notaryv1/

PR #420, opened 2023-02-27
feat(sbom): add a directive to generate SBOM for a layer

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


pkg/stacker/bom.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = SetupBuildContainerConfig(sc, storage, c, name)
+	if err != nil {
+		return err
+	}
+
+	source := "/usr/lib/"
```
smoser, line 49: this seems arbitrary, does it not?

rchincha: testing

pkg/container/container.go
```diff
@@ -138,6 +142,7 @@ func (c *Container) Execute(args string, stdin io.Reader) error {
 	// filesystem after execution. we should probably parameterize this in
 	// the storage API.
 	defer os.Remove(path.Join(c.sc.RootFSDir, c.c.Name(), "overlay", "stacker"))
+	defer os.Remove(path.Join(c.sc.RootFSDir, c.c.Name(), "overlay", "stacker-artifacts"))
```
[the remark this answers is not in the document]
rchincha: Will clean everything once this begins to work

andaaron (commented):
From user perspective I would expect the stacker command would build the sbom layer(s) in the same stacker build command, and use an additional manifest (maybe also tag it?).
Is there a spec for the SBOM artifact? 
However the test I see (and I take as an example of how to use the tool) require me as a user to execute an extra command manually inside the run().

rchincha:
2023-05-22T09:00:29.5483526Z # {"level":"error","error":"not found: /not-inventoried/file","path":"/not-inventoried/file","time":"2023-05-22T08:40:52Z","message":"inventory verify failed"}

test/bom.bats
```diff
@@ -0,0 +1,40 @@
...
+      /stacker-bom discover -o /stacker-artifacts/installed-packages.json
+      # our own custom packages
+      mkdir -p /pkg
+      touch /pkg/file
+    annotations:
+      org.opencontainers.image.authors: bom-test
+      org.opencontainers.image.vendor: bom-test
+EOF
+    stacker build
+    [ -f .stacker/artifacts/centos/installed-packages.json ]
```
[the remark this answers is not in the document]
rchincha: This should happen automatically.

cmd/stacker/main.go
```diff
@@ -29,6 +29,7 @@ var (
 )
 
 //go:embed lxc-wrapper/lxc-wrapper
+//go:embed stacker-bom
```
smoser, line 32: Is there a reason that you would stuff a go binary into another go binary?
Why not just import github.com/project-stacker/stacker-bom/pkg and use that ?

go.mod
```diff
@@ -11,41 +11,65 @@ require (
...
-	github.com/vbatts/go-mtree v0.5.2
-	golang.org/x/sys v0.6.0
-	golang.org/x/term v0.6.0
+	github.com/vbatts/go-mtree v0.5.3
+	golang.org/x/sys v0.8.0
+	golang.org/x/term v0.8.0
 	gopkg.in/yaml.v2 v2.4.0
+	sigs.k8s.io/bom v0.5.2-0.20230512052447-fef7b03b207d
 	sigs.k8s.io/yaml v1.3.0
+	stackerbuild.io/stacker-bom v0.0.0-00010101000000-000000000000
```
smoser, line 38: this is how i would have expected you to use stacker-bom.

pkg/stacker/bom.go
```diff
@@ -0,0 +1,126 @@
...
+	if sc.Debug {
+		cmd += " --debug"
+	}
+
+	cmd += " internal-go"
+
+	author := l.Annotations[types.AuthorAnnotation]
+	org := l.Annotations[types.OrgAnnotation]
+	license := l.Annotations[types.LicenseAnnotation]
+	dest := "/stacker-artifacts"
```
smoser, line 60: can we put this under /stacker ? so /stacker is read-only mount (with imports) and /stacker/artifacts is a read-write mount with this layer's artifacts.  That way we don't pollute the top level any more.

test/bom.bats
```diff
@@ -0,0 +1,111 @@
...
+        version: 1.0.0
+        license: Apache-2.0
+        paths: [/pkg1]
+      - name: pkg2
+        version: 1.0.0
+        license: Apache-2.0
+        paths: [/pkg2]
+    run: |
+      # discover installed pkgs
+      /stacker-bom discover -o /stacker-artifacts/installed-packages.json
```
smoser, line 80: I think I understand why you implemented this as a program that could be executed during 'run'. that allows the caller to run 'bom' and then remove things such as /var/lib/rpm that 'bom' would have used.

that makes sense i guess.

cmd/stacker/internal_go.go
```diff
@@ -208,3 +218,51 @@ func doAtomfsUmount(ctx *cli.Context) error {
...
+		return errors.Errorf("wrong number of args")
+	}
+
+	// merge
+	dest := ctx.Args().Get(0)
+	name := ctx.Args().Get(1)
+	author := ctx.Args().Get(2)
+	org := ctx.Args().Get(3)
+
+	if err := bom.MergeDocuments("/stacker-artifacts", name, author, org, dest); err != nil {
```
smoser, line 255: it is weird that we merge all of the things under the generically named '/stacker-artifacts' (artifacts) as if they are all 'bom').  Do we need to namespace that further, so that the directory would be /stacker-artifacts-bom ? As i mention below, i think ideally a single top level directory 'stacker' is prefereable, so /stacker/artifacts/bom ... all the files in that directory would be expected to be 'bom'.

smoser (changes requested):
I think the embedded go binary is weird. It doesn't provide much to the user, as they could just as easily have done something like this:

```yaml
import:
 - /my/stacker-bom
run: |
  #!/bin/sh
  /stacker/stacker-bom discovery -o /stacker-artifacts/stuff.json
```

I *do* like that writing bom files can be done with something other than 'stacker-bom' though (as i understand it , any files in /stacker-artifacts/*.json will be merged in).

The changes I would request are:
 * a single //stacker top level "pollution" of the container during run (i could live with 2 if we had to... maye /stacker and /stacker-data/bom )
 * drop the embed of stacker-bom, replacing it with a subcommand for stacker (stacker bom) and put stacker into the container at /stacker/stacker or something.
 * stacker-bom should "just know" to always skip /stacker (and /stacker-data) when it runs.

but overall, i think this is really nice.

cmd/stacker/internal_go.go
rchincha: Meant to catch all artifacts - bom is one, we have an inventory file also. No need to complicate the convention unnecessarily?

rchincha:
@smoser let me get an end-to-end workflow going, we can debate/address your comments after.

Makefile
```diff
@@ -34,17 +36,20 @@ stacker: $(STAGE1_STACKER) $(STACKER_DEPS) cmd/stacker/lxc-wrapper/lxc-wrapper.c
...
-stacker-dynamic: $(STACKER_DEPS) cmd/stacker/lxc-wrapper/lxc-wrapper
+stacker-dynamic: $(STACKER_DEPS) cmd/stacker/lxc-wrapper/lxc-wrapper cmd/stacker/stacker-bom
 	$(call build_stacker,,,stacker-dynamic)
 
 cmd/stacker/lxc-wrapper/lxc-wrapper: cmd/stacker/lxc-wrapper/lxc-wrapper.c
 	make -C cmd/stacker/lxc-wrapper LDFLAGS=-static LDLIBS="$(shell pkg-config --static --libs lxc) -lpthread -ldl" lxc-wrapper
 
+cmd/stacker/stacker-bom:
+	curl -Lo cmd/stacker/stacker-bom https://github.com/project-stacker/stacker-bom/releases/download/$(STACKER_BOM_VERSION)/stacker-bom-linux-amd64
+	chmod +x cmd/stacker/stacker-bom
```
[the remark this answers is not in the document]
hallyn: ping?

cmd/stacker/internal_go.go
hallyn: No need to complicate, but this is the kind of thing that once established can be painful (both to us and users) to extend later, if you decide there are other artifacts you want to deal with.  /stacker-artifacts/bom looks good to me.  (As evidence of the possible dangers, /stacker/artifacts unnecessarily risks conflicting with a user's imports)

cmd/stacker/main.go
hallyn: @rchincha was there a reason to go with go:embed here?

(Did you want to process it with m4?  I kid!)

hallyn (approved):
Looks good to me, but two questions shoudl I believe still be answered.

cmd/stacker/internal_go.go
smoser: You're right, that /stacker/artifacts does risk collision with an import from a user 'stacker.yaml' in a pre-1.0 release of stacker.  I think thats probably very low likelyhood.

The problem I have with /stacker-artifacts is that, the next time we have a feature that needs a path inside the container, you'll follow the convention and use /stacker-other-thing.

That would mean we'd have:

    /stacker
    /stacker-artifatcs
    /stacker-other-thing
    /stacker-thing-a

So if a user is to stay out of stacker's potential namespace, they have to just avoid any paths that start with /stacker.  The above just unplanned to anyone thinking about it.

I think the right thing to do is to tell users the whole tree is theirs except /stacker. 

To do that cleanly, it would make sense to then move the imports into /stacker/imports. 

It really feels like just biting the bullet and doing this is the right way forward.

rchincha: We currently only have 1) /stacker (imported), 2) /stacker-artifacts (layer related artifacts), 3) /stacker-bom (embedded binary)
In "future" expect things like in-toto attestations etc which will also land in 2)

smoser: Are you  willing to guarantee to the user that /stacker/ and /stacker-artifacts/ are the only paths that stacker will interact with?  As a user, i'd like to know that whatever (silly or problematic paths) my container has, stacker will not break or pollute my container.

As a developer, I think it is much more future-safe to adjust the top level /stacker/ now and promise that stacker will not touch any other paths. It is OK for stacker to touch paths that are explicitly requested by the user as through 'import' with a dest or other explicit mechanism.

rchincha: `import:` is a directive that can possibly collide

So here is the scheme:
/stacker/stacker-bom <- `/proc/self` bind-mounted into container
/stacker/imports/<imported> <- read-only non-dest imports (this is a breaking change)
/stacker/artifacts <- writable from inside the container

smoser: i would suggest moving 'stacker-bom' into /stacker/tools or /stacker/bin

rchincha: (revision) /stacker/tools/stacker-bom

rchincha:
2023-08-14T17:31:17.2117593Z # bind mounting /home/runner/work/stacker/stacker/stackertest-test_can_read_previous_version-27s_cache.WKvnV5/.stacker/imports/test into container
2023-08-14T17:31:17.2118033Z # + cp /stacker/imports/foo /foo
2023-08-14T17:31:17.2118442Z # cp: cannot stat '/stacker/imports/foo': No such file or directory

Since we have a breaking change in this PR

cmd/stacker/internal_go.go
```diff
@@ -208,3 +224,80 @@ func doAtomfsUmount(ctx *cli.Context) error {
...
+
+	return fs.BuildPackage(name, author, org, license, pkgname, pkgversion, paths, out)
+}
+
+func doBomVerify(ctx *cli.Context) error {
+	if ctx.Args().Len() != 4 {
+		return errors.Errorf("wrong number of args")
+	}
+
+	// merge
```
smoser, line 284: this is a strange comment here. I guess maybe it would make sense down a couple lines.  Just reading it I thought "wait, this is Verify, not merge".

Makefile
```diff
@@ -45,7 +45,6 @@ stacker-dynamic: $(STACKER_DEPS) cmd/stacker/lxc-wrapper/lxc-wrapper
 cmd/stacker/lxc-wrapper/lxc-wrapper: cmd/stacker/lxc-wrapper/lxc-wrapper.c
 	make -C cmd/stacker/lxc-wrapper LDFLAGS=-static LDLIBS="$(shell pkg-config --static --libs lxc) -lpthread -ldl" lxc-wrapper
 
-
```
smoser, line 48: just un-touch this file

cmd/stacker/internal_go.go
```diff
@@ -208,3 +224,80 @@ func doAtomfsUmount(ctx *cli.Context) error {
...
+	lic := "unknown"
+
+	if err := distro.ParsePackage(input, author, org, lic, fmt.Sprintf("/stacker/artifacts/%s.json", filepath.Base(input))); err != nil {
+		return nil
+	}
+
+	return nil
+}
+
+// pkgname, license, paths...
```
smoser, line 257: improve or remove this comment.  It doesn't line up with the Args().Get calls below.

cmd/stacker/internal_go.go
```diff
@@ -208,3 +224,80 @@ func doAtomfsUmount(ctx *cli.Context) error {
...
+	org := ctx.Args().Get(3)
+
+	if err := bom.MergeDocuments("/stacker/artifacts", name, author, org, dest); err != nil {
+		return err
+	}
+
+	// check against inventory
+	if err := fs.GenerateInventory("/",
+		[]string{"/proc", "/sys", "/dev", "/etc/resolv.conf",
+			"/stacker", "/stacker/artifacts", "/stacker-bom", "/static-stacker"},
```
smoser, line 297: i think you can drop /stacker-bom now ? and /stactic-stacker ?
this should just have '/stacker/' right?

pkg/stacker/publisher.go
```diff
@@ -65,6 +75,197 @@ func NewPublisher(opts *PublishArgs) *Publisher {
...
+	if err != nil {
+		log.Errorf("unable get digest for file:%s, err:%s", path, err)
+		return nil, err
+	}
+
+	return &dgst, nil
+}
+
+// publishArtifact to a registry/repo for this subject
+func (p *Publisher) publishArtifact(path, mtype, registry, repo, subject string, skipTLS bool) error {
```
smoser, line 128: Are there 'containers/image' library calls that you could use?

pkg/stacker/build.go
```diff
@@ -476,7 +476,7 @@ func (b *Builder) build(s types.Storage, file string) error {
 			}
 
 			// These should all be non-interactive; let's ensure that.
-			err = c.Execute("/stacker/.stacker-run.sh", nil)
+			err = c.Execute("/stacker/imports/.stacker-run.sh", nil)
```
smoser, line 479: this feels weird in /stacker/imports, but it seems unlikely to stomp on user content. so unless you think of something that fits better, this seems fine.

test/bom.bats
```diff
@@ -0,0 +1,123 @@
...
+        /etc/rpm/macros.image-language-conf /var/lib/rpm/.dbenv.lock \
+        /var/lib/rpm/Enhancename /var/lib/rpm/Filetriggername \
+        /var/lib/rpm/Recommendname /var/lib/rpm/Suggestname \
+        /var/lib/rpm/Supplementname /var/lib/rpm/Transfiletriggername \
+        /var/log/anaconda \
+        /etc/sysconfig/anaconda /etc/sysconfig/network-scripts/ifcfg-* \
+        /etc/sysconfig/sshd-permitrootlogin /root/anaconda-* /root/original-* /run/nologin \
+        /var/lib/rpm/.rpm.lock /etc/.pwd.lock /etc/BUILDTIME
+    annotations:
+      org.opencontainers.image.authors: bom-test
```
smoser, line 52: I'm curious if whitespace correctly goes through into the internal command.
Does it work if your 'authors' are "George Williams, Mary Andrews".

Same for 'vendor' being "ACME Widgets & Trinkets Inc".

pkg/stacker/build.go
```diff
@@ -706,14 +729,41 @@ func SetupLayerConfig(config types.StackerConfig, c *container.Container, l type
...
+			log.Debugf("not bind mounting %s into container", artifactsDir)
+		}
+
+		// make stacker also available to run the internal bom cmds
+		binary, err := os.Readlink("/proc/self/exe")
+		if err != nil {
+			return errors.Wrapf(err, "couldn't find executable for bind mount")
+		}
+
+		err = c.BindMount(binary, "/stacker/tools/static-stacker", "")
```
smoser, line 759: Is there a reason to call this 'static-stacker' rather than just 'stacker' ?

pkg/stacker/bom.go
```diff
@@ -0,0 +1,153 @@
...
+	binary, err := os.Readlink("/proc/self/exe")
+	if err != nil {
+		return err
+	}
+
+	if err := c.BindMount(binary, "/static-stacker", ""); err != nil {
+		return err
+	}
+
+	cmd := fmt.Sprintf("/static-stacker --oci-dir %s --roots-dir %s --stacker-dir %s --storage-type %s --internal-userns",
```
smoser, line 50: Is it possible to change this '/static-stacker' into /stacker/tools/  for consistency?

smoser (commented):
Some comments inline.
Thank you for your work on this.

pkg/stacker/publisher.go
rchincha: nope :(

test/bom.bats
rchincha: Let's amend test and try.

pkg/stacker/build.go
rchincha: Just keeping with existing convention and also run-on /stacker/tools/stacker?

pkg/stacker/build.go
rchincha: Maybe we can put this under /stacker/tools/.stacker-run.sh

pkg/stacker/bom.go
rchincha: Fixed

cmd/stacker/internal_go.go
rchincha: Fixed

Makefile
rchincha: Fixed

cmd/stacker/internal_go.go
rchincha: Fixed

pkg/stacker/build.go
rchincha: Actually, this is under "automatic" binding of /stacker/imports, let's not change that.

test/bom.bats
rchincha: test_bom_tool_should_work_inside_run.wZP2W7/oci --roots-dir /home/runner/work/stacker/stacker/stackertest-test_bom_tool_should_work_inside_run.wZP2W7/roots --stacker-dir /home/runner/work/stacker/stacker/stackertest-test_bom_tool_should_work_inside_run.wZP2W7/.stacker --storage-type overlay --internal-userns --debug internal-go bom-build /stacker/artifacts George Williams, Mary Andrews ACME Widgets & Trinkets Inc. MIT pkg1 1.0.0 /pkg1]

breaks!

pkg/stacker/build.go
smoser: i think that makes sense.

test/bom.bats
```diff
@@ -0,0 +1,125 @@
...
+        version: 1.0.0
+        license: Apache-2.0
+        paths: [/pkg1]
+      - name: pkg2
+        version: 1.0.0
+        license: Apache-2.0
+        paths: [/pkg2]
+    run: |
+      # discover installed pkgs
+      /stacker/tools/static-stacker --internal-userns internal-go bom-discover
```
smoser, line 30: The mechanism that you had before for executing stacker-bom was much cleaner.  we should not expect a user to use '--internal-userns internal-go' flags (we should never document 'internal-go' flags).

We want the documented command that the user runs to be one of:

 1. stacker bom-discover
 2. stacker-bom discover
 3. stacker bom discover

I think i prefer 3, and making 'stacker bom discover' just work wherever it was run.

I think the right way to fix that is to add the bom commands to main (not internal-go), and then make shouldSkipInternalUserns (cmd/stacker/main.go) return false for 'bom' commands.  stacker bom subcommand never needs to enter a internal user namespace. It just inspects directory trees and writes content somewhere.

cmd/stacker/internal_go.go
```diff
@@ -61,6 +65,18 @@ var internalGoCmd = cli.Command{
...
+			Name:   "bom-discover",
+			Action: doBomDiscover,
+		},
+		&cli.Command{
+			Name:   "bom-build",
+			Action: doBomBuild,
+		},
+		&cli.Command{
+			Name:   "bom-verify",
+			Action: doBomVerify,
```
smoser, line 78: We don't have any convention in stacker here yet as to whether we want:

sub-subcommands (stacker bom build) or hyphenated commands (stacker bom-build).

because we're going to expose that and even document it to the user as the way to generate bom information, we should at least think on it.

pkg/stacker/bom.go
```diff
@@ -0,0 +1,139 @@
...
+		return err
+	}
+
+	stackerInternal := []string{"/stacker/tools/static-stacker", "internal-go"}
+
+	author := l.Annotations[types.AuthorAnnotation]
+	org := l.Annotations[types.OrgAnnotation]
+	license := l.Annotations[types.LicenseAnnotation]
+	dest := "/stacker/artifacts"
+	cmd := append(stackerInternal, "bom-build", dest, author, org, license, pkg.Name, pkg.Version)
```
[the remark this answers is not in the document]
rchincha: A bit tricky - yes we should have license for each pkg entry. However, the global license is what license the container image itself is being posted under.

PR #432, opened 2023-03-06
Separate build and test in github workflows.

smoser:
Currently the "Build and test" step of the c-i workflow does both build a binary and test the binary.  Make 2 sections 'Build' and 'Test'. This way you can know which of the two sections failed, and also see how long each portion took.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


rchincha (approved):
lgtm

PR #438, opened 2023-03-07
fix(ci): add a "build-id" input var to ci builds

rchincha:
Currently, the github workflows dependencies are chained as follows:

build.yaml -> ci.yaml
release.yaml -> ci.yaml
convert.yaml -> ci.yaml

build.yaml and release.yaml are inline builds but operate on different build triggers. convert.yaml is timer-based and works offline.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -10,6 +10,10 @@ on:
         required: true
         type: string
         description: 'Stringified JSON object listing stacker privilege-level'
+      build-id:
+        required: true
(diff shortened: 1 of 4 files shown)


smoser (approved):
lgtm.
it feels like there should be a simpler way to do this. i'm surprised the checkouts don't get tags (so 'git describe' would work).
but i'm fine with this and happy to have it work.

PR #439, opened 2023-03-08
fix(ci): convert ci is failing due to perms

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/.github/workflows/convert.yaml
+++ b/.github/workflows/convert.yaml
@@ -35,10 +35,9 @@ jobs:
           git clone https://github.com/alpinelinux/docker-alpine.git
           chmod -R a+rwx docker-alpine
           cd docker-alpine
-          mkdir -p /out
-          chmod -R a+rwx /out
+          TEMPDIR=$(mktemp -d)
           stacker convert --docker-file Dockerfile --output-file stacker.yaml --substitute-file stacker-subs.yaml
-          stacker build -f stacker.yaml --substitute-file stacker-subs.yaml --substitute IMAGE=alpine
+          stacker build -f stacker.yaml --substitute-file stacker-subs.yaml --substitute IMAGE=alpine --substitute STACKER_VOL1="$TEMPDIR"
           stacker publish -f stacker.yaml --substitute-file stacker-subs.yaml --substitute IMAGE=alpine --skip-tls --url docker://${REGISTRY_URL} --layer alpine --tag latest
           rm -f stacker.yaml stacker-subs.yaml
           stacker clean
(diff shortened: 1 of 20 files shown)


rchincha:
https://github.com/project-stacker/stacker/actions/runs/4360350946/jobs/7625179702

^ pls see.

smoser:
If I understand this correctly this comes from tests/convert.bats which has `VOLUME /out` and then 'mkdir -p /out'.
Would it be an option to just change that to go to a different directory?

```bash
    cat > Dockerfile <<EOF
FROM alpine:3.16
VOLUME $HOME/out
...
```

It just seems odd to attempt that as an unprivileged user is not likely to be able to create and use a directory in /

rchincha:
> If I understand this correctly this comes from tests/convert.bats which has `VOLUME /out` and then 'mkdir -p /out'. Would it be an option to just change that to go to a different directory?
> 
> ```shell
>     cat > Dockerfile <<EOF
> FROM alpine:3.16
> VOLUME $HOME/out
> ...
> ```
> 
> It just seems odd to attempt that as an unprivileged user is not likely to be able to create and use a directory in /

We could, but `stacker convert` would have no way to know what `docker -v <outside-container-volume>:<inside-mount>` the outside-container-volume would be. This is just so that we show we can be fully automated. But one is of course welcome to mangle the output of `stacker convert` and then mangle.

smoser:
> > If I understand this correctly this comes from tests/convert.bats which has `VOLUME /out` and then 'mkdir -p /out'. Would it be an option to just change that to go to a different directory?
> > ```shell
> >     cat > Dockerfile <<EOF
> > FROM alpine:3.16
> > VOLUME $HOME/out
> > ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It just seems odd to attempt that as an unprivileged user is not likely to be able to create and use a directory in /
> 
> We could, but `stacker convert` would have no way to know what `docker -v <outside-container-volume>:<inside-mount>` the outside-container-volume would be. This is just so that we show we can be fully automated. But one is of course welcome to mangle the output of `stacker convert` and then mangle.

just give an error:
```
Docker file 'Dockerfile'  not supported.  The provided Dockerfile contains a VOLUME entry for '/out'.
/out is not writable by the current user.  Please modify the Dockerfile to contain a writable path for
the given VOLUME.
```

smoser (approved):
I think this is really good. Thanks.

The only comment i have is that I would suggest processing command line arguments in order rather than always giving precedence to '--substitute' over contents of '--substitute-file'.

That allows the caller to get whatever behavior they want (perhaps they just want to append a '--substitute-file' onto an existing command line with --substitute and have it override).

its not a big thing, due to the "static" nature of the substitute-file, so i'll approve but you're welcome to change that too.

PR #441, opened 2023-03-09
feat: report kernel version and fs type

rchincha:
stacker functionality depends on kernel versions and filesystem support. So useful to report this often, at least to debug issues.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/stacker/check.go
+++ b/pkg/stacker/check.go
@@ -1,0 +1,114 @@
+package stacker
+
+import (
+	"fmt"
+
+	"golang.org/x/sys/unix"
+)
+
+func KernelInfo() (string, error) {
+	utsname := unix.Utsname{}
+	if err := unix.Uname(&utsname); err != nil {
+		return "", err
+	}
+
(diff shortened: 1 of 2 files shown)


smoser (approved):
lgtm

pkg/stacker/check.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	st := unix.Statfs_t{}
+	if err := unix.Statfs(path, &st); err != nil {
+		return "", err
+	}
+
+	fstype, ok := fstypeMap[st.Type]
+	if !ok {
+		fstype = "unknown"
```
smoser, line 109: maybe show what the value is?

```go
fstype = fmt.Sprintf("unknown-0x%x", st.Type)
```

rchincha: return fmt.Sprintf("%s(%x)", fstype, st.Type), nil

^ so already there

PR #453, opened 2023-04-03
feat(import): copy folder contents

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/stacker/grab.go
+++ b/pkg/stacker/grab.go
@@ -11,7 +11,7 @@ import (
 )
 
 func Grab(sc types.StackerConfig, storage types.Storage, name string, source string, targetDir string,
-	mode *fs.FileMode, uid, gid int,
+	idest string, mode *fs.FileMode, uid, gid int,
 ) error {
 	c, err := container.New(sc, name)
 	if err != nil {
@@ -31,7 +31,12 @@ func Grab(sc types.StackerConfig, storage types.Storage, name string, source str
 	}
 
 	bcmd := []string{insideStaticStacker, "internal-go"}
-	err = c.Execute(append(bcmd, "cp", source, "/stacker/"+path.Base(source)), nil)
+
+	if idest == "" || source[len(source)-1:] != "/" {
+		err = c.Execute(append(bcmd, "cp", source, "/stacker/"+path.Base(source)), nil)
+	} else {
+		err = c.Execute(append(bcmd, "cp", source, "/stacker/"), nil)
+	}
 	if err != nil {
 		return err
 	}
--- a/test/import.bats
+++ b/test/import.bats
@@ -390,6 +390,79 @@ eigth:
     [ -f /dir/file2 ]
     [ -f /dir/files/file3 ]
 EOF
+    stacker build
+}
 
+@test "import with dir contents" {
+  mkdir folder1
+  touch folder1/file1
+  mkdir folder1/subfolder2
+  touch folder1/subfolder2/subfile1
+  cat > stacker.yaml <<EOF
+first:
+  from:
+    type: oci
+    url: $CENTOS_OCI
+  import:
+  - path: folder1/
+    dest: /folder1/
+  run: |
+    [ -f /folder1/file1 ]
+
+second:
+  from:
+    type: built
+    tag: first
+  run: |
+    mkdir -p /folder1
+    touch /folder1/file1
+    touch /folder1/file2
+
+third:
+  from:
+    type: oci
+    url: $CENTOS_OCI
+  import:
+    - path: stacker://second/folder1/
+      dest: /folder1/
+  run: |
+    [ -f /folder1/file1 ]
+    [ -f /folder1/file2 ]
+
+fourth:
+  from:
+    type: oci
+    url: $CENTOS_OCI
+  import:
+  - path: folder1/
+    dest: /
+  run: |
+    ls -l /
+    [ -f /file1 ]
+    mkdir -p /folder4/subfolder5
+    touch /folder4/subfolder5/subfile6
+
+fifth:
+  from:
+    type: oci
+    url: $CENTOS_OCI
+  import:
+  - path: folder1/subfolder2/
+    dest: /folder3/
+  - path: folder1/subfolder2
+    dest: /folder4/
+  - path: stacker://fourth/folder4/subfolder5/
+    dest: /folder6/
+  - path: stacker://fourth/folder4/
+    dest: /folder7/
+  run: |
+    ls -l /
+    [ -f /folder3/subfile1 ]
+    [ ! -e /folder3/subfolder2 ]
+    [ -f /folder4/subfolder2/subfile1 ]
+    [ -f /folder6/subfile6 ]
+    [ ! -e /folder6/subfolder5 ]
+    [ -f /folder7/subfolder5/subfile6 ]
+EOF
     stacker build
 }
--- a/test/import.bats
+++ b/test/import.bats
@@ -450,19 +450,70 @@ fifth:
   - path: folder1/subfolder2/
     dest: /folder3/
   - path: folder1/subfolder2
-    dest: /folder4/
+    dest: /folder4
   - path: stacker://fourth/folder4/subfolder5/
     dest: /folder6/
   - path: stacker://fourth/folder4/
     dest: /folder7/
   run: |
-    ls -l /
+    ls -l /folder*
     [ -f /folder3/subfile1 ]
     [ ! -e /folder3/subfolder2 ]
-    [ -f /folder4/subfolder2/subfile1 ]
+    [ -f /folder4/subfile1 ]
     [ -f /folder6/subfile6 ]
     [ ! -e /folder6/subfolder5 ]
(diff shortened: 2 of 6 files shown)


rchincha:
```
layer1:
  from:
    type: docker
    url: docker://ubuntu:latest
  import:
    - path: folder1/
      dest: /
  run: |
    ls /
```

^ at least this works with this PR.

Evaluating what changes will be needed.

smoser (changes requested):

test/import.bats
```diff
@@ -393,3 +393,21 @@ EOF
...
+@test "import with dir contents" {
+  mkdir folder1
+  touch folder1/file1
+  cat > stacker.yaml <<EOF
+first:
+  from:
+    type: oci
+    url: $CENTOS_OCI
+  import:
+  - path: folder1/
```
andaaron, line 406: Can you also add a test where the import path is 'stacker://...'?

rchincha: That is going to work yet, will need a part2 of this PR

test/import.bats
```diff
@@ -390,6 +390,53 @@ eigth:
...
+    dest: /
+  run: |
+    ls -l /
+    [ -f /file1 ]
+EOF
     stacker build
 }
```
andaaron, line 442: And in case of

```
import:
  path: folder1/folder2/
  dest: folder3/
```

I would expect content of folder 2 to be cooied under folder3 right? Can we add a test? Existing tests use the same folder name in src and dest, or only '/' at dest.

And a similar test case for stacker:// import.

test/import.bats
rchincha: Fixed

test/import.bats
rchincha: Fixed

andaaron (approved):
LGTM

test/import.bats
```diff
@@ -390,6 +390,79 @@ eigth:
...
+    url: $CENTOS_OCI
+  import:
+  - path: folder1/subfolder2/
+    dest: /folder3/
+  - path: folder1/subfolder2
+    dest: /folder4/
+  - path: stacker://fourth/folder4/subfolder5/
+    dest: /folder6/
+  - path: stacker://fourth/folder4/
+    dest: /folder7/
```
smoser, line 457: You are not testing an import of a 'stacker://' path without a trailing /.

You should add an import of:
```yaml
 - path: stacker://fourth/folder4
   dest: /folder8
```

And then a test of 
```bash
  [ -d /folder8/folder4 ]
```

smoser (commented):
As I understand from reading, a trailing '/' on a path that is a directory means that the contents of the provided path should be placed at 'dest', rather than inside of dest as 'dest/$(basename $path)'.

In https://github.com/project-stacker/stacker/issues/455 I suggested a different set of rules / behavior.

I  think we should just not bother with the 'basename' behavior. It makes things much more clear.

The other behavior I think we should iron out is what happens when the 'dest' exists? Does it replace it or "fill" it?

test/import.bats
rchincha: The issue is if folder8 is present then [ -d /folder8/folder4 ] will evaluate to true else false. But let me add this test and see what happens.

smoser: oh, yuck.

i think that 'target' (the full path to the imported thing) should be fully determined by the 'dest', and not dependent on either the 'path' or the contents of the filesystem to which it is being imported.

that just seems the least surprising behavior to me.

rchincha: Removed the test.

rchincha:
Documenting `docker` behavior

[Destination /folder1 doesn't exist]

1. COPY folder1 /folder1
- creates /folder1 and copies contents on folder1 -> /folder1

2. COPY folder1/ /folder1
-  same as 1.

3. COPY folder1 /folder1/
- same as 1.

4. COPY folder1/ /folder1/
- same as 1.

smoser:
ADD definitely looks like a bunch of iterative behaviors. complete with behavior like:

> If <src> is a local tar archive in a recognized compression format (identity, gzip, bzip2 or xz) then it is unpacked as a directory. 

Also, ADD seems to indicate *add* not "merge", and I really could be missing something in the long description of ADD's behavior, but I don't see what happens if I ADD directory 'foo' to /etc which already has 'foo'.  Do the contents get merged somehow? As I can tell, that is "union" and I only see that term used wrt tar.

if you want to use that as your reference of something to implement, and then attempt to keep your implementation bug-for-bug in sync with docker's then I guess its fine.   I just think that it is more complex than required.

rchincha:
> ADD definitely looks like a bunch of iterative behaviors. complete with behavior like:
> 
> > If  is a local tar archive in a recognized compression format (identity, gzip, bzip2 or xz) then it is unpacked as a directory.
> 
> Also, ADD seems to indicate _add_ not "merge", and I really could be missing something in the long description of ADD's behavior, but I don't see what happens if I ADD directory 'foo' to /etc which already has 'foo'. Do the contents get merged somehow? As I can tell, that is "union" and I only see that term used wrt tar.
> 
> if you want to use that as your reference of something to implement, and then attempt to keep your implementation bug-for-bug in sync with docker's then I guess its fine. I just think that it is more complex than required.

Not yet sure what stacker does in all these scenarios to compare/contrast yet.

Not one-on-one, but we are constrained by
1. stacker cannot deviate too much from docker behavior, but it may not have to
2. stacker imports (and copies) are set up as overlay-dirs

rchincha:
> but I don't see what happens if I ADD directory 'foo' to /etc which already has 'foo'. Do the contents get merged somehow? As I can tell, that is "union" and I only see that term used wrt tar.

It is a union. If the entry exists in the destination, it gets overwritten.

test/import.bats
```diff
@@ -390,6 +390,79 @@ eigth:
...
+first:
+  from:
+    type: oci
+    url: $CENTOS_OCI
+  import:
+  - path: folder1/
+    dest: /folder1/
+  run: |
+    [ -f /folder1/file1 ]
+
```
smoser, line 411: If I'm reading things correctly, then these two snippets do the same thing:

 *
   ```yaml
   import:
    - path: folder1
      dest: /
   ```

and

 *
   ```yaml
   import:
      - path: folder1
        dest: /folder1/
   ```

I find it odd that the trailing '/' on a 'path' element only means something if the 'dest' is /.

test/convert.bats
```diff
@@ -48,7 +48,6 @@ EOF
 }
 
 @test "alpine" {
-  skip_slow_test
```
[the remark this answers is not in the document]
smoser: > was this set of changes intentional?

I see that this is part of "test - remove commit", so maybe remove that commit?

test/import.bats
rchincha: It means that folder1 is a dir and everything in src goes under folder1/ as a child unambiguously.

test/convert.bats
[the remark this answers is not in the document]
rchincha: Will do. Once I am convinced that `stacker convert` tests are not broken with these changes.

pkg/stacker/grab.go
```diff
@@ -31,7 +31,12 @@ func Grab(sc types.StackerConfig, storage types.Storage, name string, source str
 	}
 
 	bcmd := []string{insideStaticStacker, "internal-go"}
-	err = c.Execute(append(bcmd, "cp", source, "/stacker/"+path.Base(source)), nil)
+
+	if idest == "" || source[len(source)-1:] != "/" {
+		err = c.Execute(append(bcmd, "cp", source, "/stacker/"+path.Base(source)), nil)
+	} else {
+		err = c.Execute(append(bcmd, "cp", source, "/stacker/"), nil)
```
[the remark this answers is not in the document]
rchincha: line #35 
if idest == ""

test/import.bats
```diff
@@ -390,6 +390,79 @@ eigth:
...
+second:
+  from:
+    type: built
+    tag: first
+  run: |
+    mkdir -p /folder1
+    touch /folder1/file1
```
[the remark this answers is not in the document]
rchincha: Yup.

test/import.bats
hallyn: I still think that rsync semantics make the most sense.

rsync -va src/ dest/ will copy contents of src/ into dest/.
rsync -va src dest/ will copy src as a directory under dest/ (plus of course all its contents).

test/import.bats
```diff
@@ -450,19 +450,70 @@ fifth:
...
+  cat > stacker.yaml <<EOF
+src_folder_dest_non_existent_folder_case1:
+  from:
+    type: docker
+    url: docker://ubuntu:latest
+  import:
+  - path: folder1
+    dest: /folder2
+  run: |
+    [ -f /folder2/file1 ]
```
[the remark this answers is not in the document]
hallyn: Mind you I think I like the behavior you have here better.  I'm surprised by what rsync is doing.

So I'm not asking you to change it.  Just document it :)

hallyn (approved):

PR #457, opened 2023-04-20
refactor!: drop support for '->' artwork

rchincha:
BREAKING CHANGE: '->' artwork syntax is dropped so older "bind:" directives will break, so users need to migrate/fix.

stacker syntax uses a strange mix of yaml and artwork such as '->', for example in "bind:" directive. This PR makes it normal declarative yaml.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


mikemccracken:
I support leaving the ascii art behind, but if you can add a deprecation warning that explains how to replace it, that'd be really helpful. 

people currently using stacker will hit this in some number of months, when someone else updates their tool version, and they will need help understanding how to fix it.

rchincha:
> I support leaving the ascii art behind, but if you can add a deprecation warning that explains how to replace it, that'd be really helpful.
> 
> people currently using stacker will hit this in some number of months, when someone else updates their tool version, and they will need help understanding how to fix it.

Yup. Like so: https://github.com/project-stacker/stacker/pull/457/files#diff-2f7f69286606c0ff1642fac8ecb40ee91874b65bb544af172e32552c91a44799R81

PR #459, opened 2023-04-21
Fix link failure by adding zstd-static to the alpine build container.

smoser:
Top commit (54b6501b188041019c19dd6d93c66838bae6c165) on alpine zstd tree commit says:

    main/zstd: split -static

That broke our build as the new zstd-dev does not contain a libzstd.a.

Thankfully the fix is as easy as adding zstd-static to the list of alpine packages in the container.


rchincha (approved):
lgtm

PR #461, opened 2023-04-25
Feature/no more anuvu/squashfs dep

smoser:
stacker was only using anuvu/squashfs import to get the size of the squashfs file (from its superblock, rather than from 'stat').

That go library used libsquashfs-tools-ng as a c-go dependency, which meant a build and runtime dependency on it.

The change here is to drop that import and instead use some native go.

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -37,7 +37,7 @@ packages:
 #### **Ubuntu 22.04**
 
     sudo apt install lxc-dev libacl1-dev libgpgme-dev libcap-dev libseccomp-dev
-    sudo apt install libpam0g-dev libselinux-dev libssl-dev libzstd-dev libcryptsetup-dev libdevmapper-dev cryptsetup-bin pkg-config libsquashfs1 libsquashfs-dev
+    sudo apt install libpam0g-dev libselinux-dev libssl-dev libzstd-dev libcryptsetup-dev libdevmapper-dev cryptsetup-bin pkg-config
 
 
 **To run `make check` you will also need:**
@@ -45,18 +45,6 @@ packages:
     sudo apt install bats jq tree
 
 **umoci** - https://github.com/opencontainers/umoci
-
-**squashtool**, but with a slightly different config than what is mentioned in the install guide (see below) - https://github.com/anuvu/squashfs
-
-Contrary to what the documentation in squashfs implies, squashtool and
-libsquash from squash-tools-ng need to be installed globally, as user specific
-path overrides aren't propagated into `make check`'s test envs.
-
-Thus, when you reach the step **install into mylocal="$HOME/lib"** from the squashfs guide, use the config below. You can put them at the end of your .bashrc file so you don't need to run them every time.
-
-    mylocal="/usr/local"
-    export LD_LIBRARY_PATH=$mylocal/lib${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}
-    export PKG_CONFIG_PATH=$mylocal/lib/pkgconfig${PKG_CONFIG_PATH:+:$PKG_CONFIG_PATH}
 
 Since the path **/usr/local** is owned by root, when you reach the step to run **make install**, you need to run it as **sudo**.
 
--- a/test/caching.bats
+++ b/test/caching.bats
@@ -185,6 +185,7 @@ EOF
 }
 
 @test "can read previous version's cache" {
+    skip "old stacker will not build because libsquashfs1-dev is not installed"
     git clone https://github.com/project-stacker/stacker
     (cd stacker && make LXC_BRANCH=$(grep ^LXC_BRANCH Makefile | cut -d'=' -f 2) LXC_CLONE_URL=$LXC_CLONE_URL)
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -53,8 +53,6 @@ jobs:
           GO111MODULE=off go get github.com/opencontainers/umoci/cmd/umoci
           curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin
           sudo apt-get install -yy autoconf automake make autogen autoconf libtool binutils git squashfs-tools libcryptsetup-dev libdevmapper-dev cryptsetup-bin squashfuse
-          (cd /tmp && git clone https://github.com/AgentD/squashfs-tools-ng && cd squashfs-tools-ng && ./autogen.sh && ./configure --prefix=/usr && make -j2 && sudo make -j2 install && sudo ldconfig -v)
-          (cd /tmp && git clone https://github.com/anuvu/squashfs && cd squashfs && make && sudo cp squashtool/squashtool /usr/bin)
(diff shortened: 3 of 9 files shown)


smoser:
Given the superblock.go is copied from diskfs, I wonder if I need to add something like this:

```diff
diff --git a/pkg/squashfs/superblock.go b/pkg/squashfs/superblock.go
index a90ab5b..c16e1ee 100644
--- a/pkg/squashfs/superblock.go
+++ b/pkg/squashfs/superblock.go
@@ -1,3 +1,12 @@
+/*
+Copyright (c) 2017 Avi Deitcher
+
+This file was initially copied from go-diskfs [1].
+The copied portion is licensed under the terms of the MIT license [2].
+ [1] https://github.com/diskfs/go-diskfs/filesystem/squashfs/superblock.go
+ [2] https://opensource.org/licenses/MI
+*/
+
 package squashfs
 
 import (

```

tych0:
My only thought is that it would be nice to import superblock.go instead of copying it, but I suppose you tried that and it introduces some other C deps that you have to build?

smoser:
The next trick to drop dependencies would be to just "shell out" to cryptsetup and dmsetup.
Normally I would argue heavily against shelling out.  But I think it is different here for a couple reasons:
 * 'dmsetup' and 'cryptsetup' seem to be reasonable dependencies to me . They are very mature cli interfaces that are very standard on linux.
 * dmsetup and cryptsetup cover the vast majority of the dynamic dependencies that stacker needs to use.

when atomfs builds, it gets a dependency (from ldd) like the following. I marked the ones that come from cryptsetup with a '-'.

```
 /lib64/ld-linux-x86-64.so.2 (0x00007f8a04774000)
-libargon2.so.1 => /lib/x86_64-linux-gnu/libargon2.so.1
-libblkid.so.1 => /lib/x86_64-linux-gnu/libblkid.so.1
-libcrypto.so.3 => /lib/x86_64-linux-gnu/libcrypto.so.3
-libcryptsetup.so.12 => /lib/x86_64-linux-gnu/libcryptsetup.so.12
 libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6
-libdevmapper.so.1.02.1 => /lib/x86_64-linux-gnu/libdevmapper.so.1.02.1
-libjson-c.so.5 => /lib/x86_64-linux-gnu/libjson-c.so.5
 liblz4.so.1 => /lib/x86_64-linux-gnu/liblz4.so.1
 liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5 
-libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 
-libpcre2-8.so.0 => /lib/x86_64-linux-gnu/libpcre2-8.so.0 
-libselinux.so.1 => /lib/x86_64-linux-gnu/libselinux.so.1 
-libudev.so.1 => /lib/x86_64-linux-gnu/libudev.so.1 
-libuuid.so.1 => /lib/x86_64-linux-gnu/libuuid.so.1 
 libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 
 libzstd.so.1 => /lib/x86_64-linux-gnu/libzstd.so.1 
 linux-vdso.so.1 
```

So basically if we shelled out, i think we could maybe get atomfs's deps down to only those compression libraries.

(note, stacker probably has others, i was just looking at atomfs now, and it
only imports stackerbuild.io/stacker/pkg/atomfs and
stackerbuild.io/stacker/pkg/mount, but a lot of functionality comes from just that)

smoser:
> My only thought is that it would be nice to import superblock.go instead of copying it, but I suppose you tried that and it introduces some other C deps that you have to build?

All diskfs exposes is a [Filesystem](https://github.com/diskfs/go-diskfs/blob/e29f8c4774c504fb5166defa77214e499e6e9875/filesystem/filesystem.go#L10) interface.  You do a [Read](https://github.com/diskfs/go-diskfs/blob/e29f8c4774c504fb5166defa77214e499e6e9875/filesystem/squashfs/squashfs.go#L100) and get one back, but then you don't have any way to get 'size' or anything even though they're in the struct.

I could propose an upstream change to expose them in a consistent manner for each of the Filesystem interface implementations; this is just lazy.

tych0:
You could downcast to the right one if the individual implentations are public.

smoser:
> You could downcast to the right one if the individual implentations are public.
> […](#)

[squashfs.Filesystem](https://github.com/diskfs/go-diskfs/blob/e29f8c4774c504fb5166defa77214e499e6e9875/filesystem/squashfs/squashfs.go#L23) is what you get back from the Read.   It has a '[superblock]' of type [*superblock](https://github.com/diskfs/go-diskfs/blob/e29f8c4774c504fb5166defa77214e499e6e9875/filesystem/squashfs/superblock.go#L65) member that is not exposed (lower case).  The superblock type has a non-exposed uint64 size.  

Maybe i could manage to get it through an import, but not in any way that was clean or imo preferable to this one file copy.

smoser:
oh for pete's sake. test taht failed is tests/caching.bats "can read previous version's cache".
That test does a clone of stacker and attempts to build it.  I can't build the previous version of stacker because I don't have one of it's dependencies (libsquashfs1 -- thats what i'm tryint to fix).

tych0:
Yeah, you'd have to do reflection, and then you're at risk of some non-public API change. Another option would be to fork + go replace, but copy seems fine too.

smoser:
I think 

> The next trick to drop dependencies would be to just "shell out" to cryptsetup and dmsetup. Normally I would argue heavily against shelling out. But I think it is different here for a couple reasons:
...

I guess there are the following cases
1. kernel 'mount' of a squashfs+dmverity image
2. creation of verity data and root hash

It seems that mount is pretty straightforward with dmsetup or cryptsetup. creation might be a little more difficult.
if we separated use and creation into separate directories, then a consumer (such as 'machine' or 'atomfs') that only needed mount would not end up with the library dep.

test/caching.bats
```diff
@@ -185,6 +185,7 @@ EOF
 }
 
 @test "can read previous version's cache" {
+    skip "old stacker will not build because libsquashfs1-dev is not installed"
```
rchincha, line 188: No, we don't want to disable this in general right?
We expect this test alone to fail and we can force-merge.

doc/install.md
```diff
@@ -37,7 +37,7 @@ packages:
 #### **Ubuntu 22.04**
 
     sudo apt install lxc-dev libacl1-dev libgpgme-dev libcap-dev libseccomp-dev
-    sudo apt install libpam0g-dev libselinux-dev libssl-dev libzstd-dev libcryptsetup-dev libdevmapper-dev cryptsetup-bin pkg-config libsquashfs1 libsquashfs-dev
+    sudo apt install libpam0g-dev libselinux-dev libssl-dev libzstd-dev libcryptsetup-dev libdevmapper-dev cryptsetup-bin pkg-config
```
rchincha, line 40: nice!

test/caching.bats
rchincha: The intent I believe is PRs don't break backward compatibility. We can disable this test for now, but we need a new PR to re-enable it right afterwards.

rchincha (approved):
lgtm

PR #463, opened 2023-04-28
fix: update and re-enable test "test can read previous version's cache"

smoser:
Commit 9671d8f35986 disabled the caching.bats test "test can read previous version's cache".

This re-enables that test after updating it to use a specific value (v1.0.0-rc4) as the "previous version".

The change to use a specific version and use its released binary improves the test.  That improves the test for the following reasons:

 * This should be faster - we don't have to build stacker twice.
 * Ultimately testing against the last released version is probably what you want... that is the cache that your users have, not the last development commit.
 * Building stacker from another version is both slow and hard to guarantee. This removes the assumption that the current build environment can build another version of stacker.
 * There was assumed knowledge about both the current and other stacker build systems in the 'make LXC_BRANCH=$(grep ...)' command.
 * The prior version of the test used "current/main" as "previous", but current/main is a moving target.  Re-running a checkout at a future point in time may fail as current/main has moved forward. (Think about using 'git bisect' to diagnose a past regression)
 * Even when working as intended, a series of PRs could still break re-use of the cache.  Given PRs A, B, C that all are merged to master.  B is tested against A, C is tested against B, but nothing tested C against A.

The change here *does* mean the following things are true, which are possibly undesirable:

 * There is a requirement to update the old version string. v1.0.0-rc4 is an arbitrary value.
 * The old binary must run in the current build environment.  This may be hard to guarantee.  I'm honestly not sure if static binary will be more reliable than building and executing a specific old version of source.

### DIFF
--- a/test/caching.bats
+++ b/test/caching.bats
@@ -185,13 +185,19 @@ EOF
 }
 
 @test "can read previous version's cache" {
-    skip "old stacker will not build because libsquashfs1-dev is not installed"
-    git clone https://github.com/project-stacker/stacker
-    (cd stacker && make LXC_BRANCH=$(grep ^LXC_BRANCH Makefile | cut -d'=' -f 2) LXC_CLONE_URL=$LXC_CLONE_URL)
-
     # some additional testing that the cache can be read by older versions of
     # stacker (cache_test.go has the full test for the type, this just checks
     # the mechanics of filepaths and such)
+    local relurl="https://github.com/project-stacker/stacker/releases/download"
+    local oldver="v1.0.0-rc4"
+    local oldbin="./stacker-$oldver"
+    if [ "$PRIVILEGE_LEVEL" != "priv" ]; then
+        skip_if_no_unpriv_overlay
+    fi
+
+    wget -O "$oldbin" --progress=dot:mega "$relurl/$oldver/stacker"
+    chmod 755 "$oldbin"
+
     touch foo
     cat > stacker.yaml <<EOF
 test:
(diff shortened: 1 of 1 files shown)


smoser:
https://github.com/project-stacker/stacker/pull/461 disabled the test because it would fail.

hallyn (approved):

test/caching.bats
```diff
@@ -185,13 +185,19 @@ EOF
...
 @test "can read previous version's cache" {
-    skip "old stacker will not build because libsquashfs1-dev is not installed"
-    git clone https://github.com/project-stacker/stacker
-    (cd stacker && make LXC_BRANCH=$(grep ^LXC_BRANCH Makefile | cut -d'=' -f 2) LXC_CLONE_URL=$LXC_CLONE_URL)
-
     # some additional testing that the cache can be read by older versions of
     # stacker (cache_test.go has the full test for the type, this just checks
     # the mechanics of filepaths and such)
+    local relurl="https://github.com/project-stacker/stacker/releases/download"
+    local oldver="v1.0.0-rc4"
```
rchincha, line 192: oldver=$(git ls-remote --refs --sort="version:refname" --tags $repo | cut -d/ -f3-|tail -n1)

should do it?

rchincha: oldver=$(git describe --abbrev=0 --tags)

smoser: that would still result in the shortcoming:

>  * Even when working as intended, a series of PRs could still break
>    re-use of the cache.  Given PRs A, B, C that all are merged to
>    master.  B is tested against A, C is tested against B, but
>    nothing tested C against A.

it wont stop release N+1 from regressing release if there were any tags in between.

Issue #456, opened 2023-04-18
Bug: import when path is a (http/https) url does not allow renaming


smoser:
### stacker version

v1.0.0-rc4-8e267fc

### Describe the bug

stacker cannot rename an import when the 'path' is a http or https url.
the imported path will be the basename of the url.

### To reproduce

1. Configuration
stacker.yaml
```yaml
build:
  build_only: true
  from: 
    type: docker
    url: ${{DOCKER_BASE:docker://docker.io/library}}/busybox
  import:
    - path: /etc/passwd
      dest: /imports/host-file
    - path: https://www.cisco.com/favicon.ico
      dest: /imports/http-file
  run: |
    #!/bin/sh
    set +e
    
    ERRS=0
    err() { echo "$@"; ERRS=$((ERRS+1)); }
    assertFile() {
      [ -f "$1" ] || {
        err "$1: not a file"
        return 1
      }
      return 0
    }

    ls -l /imports
    assertFile /imports/host-file
    assertFile /imports/http-file

    [ "$ERRS" = "0" ]

```

3. Client tool used: stacker
4. Seen error

rchincha:
- path: https://www.cisco.com/favicon.ico
      dest: /imports/http-file

smoser:
> ```
>  - path: https://www.cisco.com/favicon.ico
>    dest: /imports/http-file
> ```

thats what i had, right?

```yaml
    - path: https://www.cisco.com/favicon.ico
      dest: /imports/http-file
```

I'm confused.

PR #466, opened 2023-05-01
fix: support dest path for import-http

rchincha:
Fixes issue #456

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/test/import-http.bats
+++ b/test/import-http.bats
@@ -73,4 +73,23 @@ function teardown() {
     [ "$(sha reference/nm_orig)" == "$(sha dest/img/rootfs/root/nm)" ]
 }
 
+@test "importing to a dest" {
+    cat > img/stacker1.yaml <<EOF
+centos_base:
+    from:
+        type: oci
+        url: $CENTOS_OCI
+    import:
+        - path: https://www.cisco.com/favicon.ico
+          dest: /dest/icon
+    run: |
+        [ -f /dest/icon ]
+        [ ! -f /dest/favicon.ico ]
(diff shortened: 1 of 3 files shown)


test/import-http.bats
```diff
@@ -73,4 +73,34 @@ function teardown() {
...
+    run: |
+        ls
+EOF
+    cat > img/stacker2.yaml <<EOF
+img:
+    from:
+        type: oci
+        url: $(pwd)/oci:centos_base
+    import:
+        - path: http://network-test.debian.org/nm
```
smoser, line 91: Could you just re-use `https://www.cisco.com/favicon.ico` that is used in several other places in tests ? that way we dont' needlessly extend the infrastructure that tests depend on.

test/import-http.bats
```diff
@@ -73,4 +73,34 @@ function teardown() {
     [ "$(sha reference/nm_orig)" == "$(sha dest/img/rootfs/root/nm)" ]
 }
 
+@test "importing to a dest" {
+    cat > img/stacker1.yaml <<EOF
```
smoser, line 77: what do you gain by building 'stacker1' and 'stacker2' here rather than just one  thing?

test/import-http.bats
rchincha: Ok, done

test/import-http.bats
rchincha: Was putting together a quick test. Fixed now.

smoser (approved):
this good to me. thanks.

PR #474, opened 2023-07-05
fix: added build-base to package list for static build

ashwing123:
**What type of PR is this?**
bug

**Which issue does this PR fix**:
fixes #473

**What does this PR do / Why do we need it**:
Enables alpine:edge aarch64 image to finish building (allows it to compile C programs).

**If an issue # is not available please add repro steps and logs showing the issue**:

**Testing done on this change**:

**Automation added to e2e**:

**Will this break upgrades or downgrades?**

**Does this PR introduce any user-facing change?**:

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Signed-off-by: Ashwin Gopalan ashwin.gopalan02@gmail.com

PR #469, opened 2023-06-07
feat: build for other os/arch

rchincha:
**What type of PR is this?**

**Which issue does this PR fix**:

**What does this PR do / Why do we need it**:

**If an issue # is not available please add repro steps and logs showing the issue**:

**Testing done on this change**:

**Automation added to e2e**:

**Will this break upgrades or downgrades?**

**Does this PR introduce any user-facing change?**:

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #475, opened 2023-07-11
feat: adding build-base on top of rchincha's arm additions

ashwing123:
**What type of PR is this?**
feature (adding onto #469)

See #474 for reasoning behind adding build-base. This is just to test its compatibility with Ram's arm additions/workflow build.

**Testing**
Tested `build-base` addition on ampere arm box and it enabled stacker to build.


hallyn:
Not quite sure what's going on here with the set of commits :)  Please ping me out of band when it's ready for review.

PR #481, opened 2023-08-16
Execute inside container through new internal command exec-args.

smoser:
Stacker executes commands inside a container by setting the lxc config name 'lxc.execute.cmd'.  That is a scalar value but we're ultimately hoping to execute an array.
LXC does support quote escaping in the value, but it does not support nested quotes and other characters could be an issue.

The change here is to add an internal command 'exec-args' and then always utilize that.  exec-args takes a single argument which is expected to be a base64 encoded json string array.

exec-args just decodes the base64 content and then unmarshals a json string array and executes the result.

As an example use:

    $ cat b64blob
    #!/usr/bin/python3
    import json, base64, sys
    print(base64.b64encode(json.dumps(sys.argv[1:]).encode()).decode())

    $ blob=$(./b64blob sh -c \
       'i=0; while i=$((i+1)) && [ $# -ne 0 ]; do \
           printf "%02d: %s$\n" "$i" "$1"; shift; done' \
           -- 'arg one !' "two tooo  " ' -> "thi$;" is three'

    $ ./stacker internal-go exec-args $blob
    01: arg one !$
    02: two tooo  $
    03:  -> "thi$;" is three$

I even validated that an emoji correctly gets through, but didn't want put it in a commit message.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -2,6 +2,8 @@ package container
 
 import (
 	"bufio"
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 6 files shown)


pkg/container/container.go
```diff
@@ -118,8 +120,18 @@ func (c *Container) containerError(theErr error, msg string) error {
...
 
-func (c *Container) Execute(args string, stdin io.Reader) error {
-	if err := c.SetConfig("lxc.execute.cmd", args); err != nil {
+func (c *Container) Execute(args []string, stdin io.Reader) error {
+	jblob, err := json.Marshal(args)
+	if err != nil {
+		return err
+	}
+
+	base64.StdEncoding.EncodeToString([]byte(jblob))
```
hallyn, line 129: I'm confused.  This line does nothing, right?

smoser: correct. probably left over.

PR #486, opened 2023-08-21
fix: handle error else leads to a nil ptr access

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


pkg/lib/file.go
```diff
@@ -83,6 +83,11 @@ func FindFiles(base, pattern string) ([]string, error) {
 
 	visit := func(path string, info os.FileInfo, err error) error {
 
+		if err != nil {
+			// ignore errors and continue
+			return nil
```
hallyn, line 88: The library documentation instead shows a 'return err' in this case.  Can you explain why you are returning nil?

rchincha: https://pkg.go.dev/path/filepath#WalkFunc

The error result returned by the function controls how Walk continues. If the function returns the special value SkipDir, Walk skips the current directory (path if info.IsDir() is true, otherwise path's parent directory). If the function returns the special value SkipAll, Walk skips all remaining files and directories. Otherwise, if the function returns a non-nil error, Walk stops entirely and returns that error.

rchincha: Do you want to stop the walk due to errors - permissions, etc on unrelated files potentially?

smoser: > Do you want to stop the walk due to errors - permissions, etc on unrelated files potentially?

isn't it just swallowing errors otherwise?

rchincha: OTOH, do you want to stop looking for files for such errors?

hallyn: Wait - github ate my reply??  How annoying.

hallyn: @smoser if we are looking for x/steve, but x/bob is 000 perms and causes us an error on lookup, then we don't want to stop looking for steve bc of that.

So ignoring the error is right I think, in this case.

hallyn (approved):

pkg/lib/file.go
smoser: how do you know the error that was thrown away is one that you would have ignored?

smoser: As an example that is relevant here:

```
$ mkdir -p  layers/{layer1,layer2}
$ touch layers/{layer1,layer2}/stacker.yaml
$ chmod 750  layers
$ sudo chown root:root layers
$ ls -l layers/layer1
ls: cannot access 'layers/layer1': Permission denied
$ stacker --debug recursive-build --search-dir=layers
stacker version v1.0.0-rc4-8e267fc
usernsexec-ing [u 0 1000 1 1 100001 65535 g 0 1000 1 1 100001 65535 -- /usr/local/bin/s
tacker --internal-userns --debug recursive-build --search-dir=layers]
stacker version v1.0.0-rc4-8e267fc
stacker build order:
$ echo $?
0
```

As it is, current master behaves badly on this. I told it to look in layers/ for my stacker.yaml files. it couldn't read layers, surprisingly did nothing, and then exited success.

i think this change makes for more surprising / unexpected errors that are not easy to fix.

If you get "permission denied 'layer/'" errors, then you have an idea on whats wrong.  If stacker just does nothing, then your c-i thinks it worked or you end up using old layers that you thought you just built or something else.

hallyn: Ok, yeah, let's fail on error, and user can fix the situation.

rchincha: 2023-08-22T21:33:57.8999705Z   FindFiles ✔✔✘
2023-08-22T21:33:57.8999838Z 
2023-08-22T21:33:57.8999844Z 
2023-08-22T21:33:57.8999931Z Failures:
2023-08-22T21:33:57.9000054Z 
2023-08-22T21:33:57.9000250Z   * stackerbuild.io/stacker/pkg/lib/file_test.go 
2023-08-22T21:33:57.9000512Z   Line 50:
2023-08-22T21:33:57.9000711Z   Expected: nil
2023-08-22T21:33:57.9001116Z   Actual:   'open /tmp/snap-private-tmp: permission denied'
2023-08-22T21:33:57.9001319Z 

^ not really right?

rchincha: 43     Convey("FindFiles", t, func() {
 44         src, err := os.CreateTemp("", "src")
 45         So(err, ShouldBeNil)
 46         So(src, ShouldNotBeNil)
 47         defer os.Remove(src.Name())
 48
 49         files, err := lib.FindFiles(path.Dir(src.Name()), ".*")
 50         So(err, ShouldBeNil)
 51         So(files, ShouldNotBeEmpty)
 52     })

hallyn (approved):

PR #489, opened 2023-08-22
fix: Move bom to top level subcmd, simplify shouldSkipInternalUserns

smoser:
The bom commands are useful outside of stacker, and since they're exposed to the user, they should *not* be part of the hidden internal-go interface.

So, change:

 - stacker internal-go bom-discover
 + stacker bom discover

Also, shouldSkipInternalUserns was overly complicated as a result of iterative development.  Its simple now and based only on the subcommand name.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


test/bom.bats
```diff
@@ -27,7 +27,7 @@ bom-parent:
         paths: [/pkg2]
     run: |
       # discover installed pkgs
-      /stacker/tools/static-stacker internal-go bom-discover
```
rchincha, line 30: actually ok with this change

hallyn (approved):

PR #493, opened 2023-08-30
fix: only use workaround for overlay if there are zero lowerdirs

smoser:
A comment in metadata.go says that

  > overlayfs doesn't work with < 2 lowerdirs, so we add some

That doesn't seem right, you definitely do need 1 lowerdir, but you do not need 2.


rchincha (approved):
lgtm

mikemccracken (approved):

PR #494, opened 2023-08-30
fix: clarify comments in molecule.Mount and lxcRootfsString

smoser:
fix: clarify comments in molecule.Mount and lxcRootfsString

molecule.overlayArgs provides options for a kernel mount-option string.
Kernel mount options for overlay require *more* than one lowerdir if
there is no upperdir.  molecule.Mount always performs read-only
mounts (thus no upperdir). So molecule.overlayArgs must add
a workaround lowerdir if there is only one.

Separately, lxcRootfsString function returns a string to be supplied to
lxc's lxc.rootfs.path.  lxc.rootfs.path is of the form:
  overlayfs:lowerdir[:lowerdir2:lowerdir3]:upperdir.
It requires 1 or more lowerdirs and an upperdir.
So in lxcRootfsString we need to create a workaround dir only if there
are zero lowerdirs (an empty container).

this is a follow-on to https://github.com/project-stacker/stacker/pull/493

### DIFF
--- a/pkg/atomfs/molecule.go
+++ b/pkg/atomfs/molecule.go
@@ -65,8 +65,9 @@ func (m Molecule) mountUnderlyingAtoms() error {
 	return nil
 }
 
-// overlayArgs returns all of the mount options to pass to the kernel to
+// overlayArgs - returns all of the mount options to pass to the kernel to
 // actually mount this molecule.
+// This function assumes read-only. It does not provide upperdir or workdir.
 func (m Molecule) overlayArgs(dest string) (string, error) {
 	dirs := []string{}
 	for _, a := range m.Atoms {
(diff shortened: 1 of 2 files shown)


hallyn:
The comment is true if there are no upperdirs.  I couldn't easily tell with quick code inspection - do you know whether this path can happen with no upperdir?

mikemccracken:
moved from chat to avoid getting it lost:

 I think the comment is wrong, but isn't it possible for there to be a molecule with just one atom?
and hence one dir, requiring a fake workaround lowerdir...

smoser:
OK, Sorry for the confusion and for taking you all down the rathole with me.

Now this PR is a comment-only change.  Hopefully it commit message and the comments changes are an improvement.

mikemccracken (approved):

hallyn (approved):

PR #497, opened 2023-09-06
Fix/nightly ci

smoser:
This is an attempt at fixing nightly c-i similar to https://github.com/project-stacker/stacker/pull/496
but with more re-use and building more nightkly.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -1,34 +1,10 @@
-name: "Nightly"
+name: "nightly"
 
 on:
   workflow_dispatch:
   schedule:
     - cron: '0 0 * * *'
 
-env:
-  REGISTRY_URL: localhost:5000
-
 jobs:
-  slow-tests:
-    name: "slow tests"
-    runs-on: ubuntu-20.04
-    steps:
-      - uses: actions/cache@v3
-        id: restore-build
-        with:
-          path: stacker
-          key: ${{ github.sha }}
-      - name: zot registry
-        env:
-          ZOT_HOST: localhost
-          ZOT_PORT: 5000
-        run: |
-          # start a zot instance (minimal)
-          podman run -d -p ${ZOT_PORT}:${ZOT_PORT} ghcr.io/project-zot/zot-minimal-linux-amd64:latest
-          # check if reachable
-          while true; do x=0; curl -f http://${REGISTRY_URL}/v2/ || x=1; if [ $x -eq 0 ]; then break; fi; sleep 1; done
-      - name: test
-        env:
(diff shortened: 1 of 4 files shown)


hallyn:
Seems reasonable.

smoser:
So it looks like this doesn't work as intended.

I made the nightly workflow run on PR in the top most commit.
 * [nightly](https://github.com/project-stacker/stacker/actions/runs/6097394094/job/16544941059?pr=497)
 * [c-i](https://github.com/project-stacker/stacker/actions/runs/6097394092/workflow?pr=497)

Both of those show in their test output:
```
ok 19 all container contents must be accounted for # skip test_all_container_contents_must_be_accounted_for is slow. Set SLOW_TEST=true to run.
ok 20 bom tool should work inside run # skip test_bom_tool_should_work_inside_run is slow. Set SLOW_TEST=true to run.
```

Further investigation shows that I had SLOW_TESTS (plural) in the workflow 'env' while the code is looking at SLOW_TEST (singular).

Change made, pushed, lets see.

After this runs i have to still drop the top-most commit so that nightly doesn't run on PR.

.github/workflows/nightly.yaml
```diff
@@ -4,31 +4,10 @@ on:
...
-          podman run -d -p ${ZOT_PORT}:${ZOT_PORT} ghcr.io/project-zot/zot-minimal-linux-amd64:latest
-          # check if reachable
-          while true; do x=0; curl -f http://${REGISTRY_URL}/v2/ || x=1; if [ $x -eq 0 ]; then break; fi; sleep 1; done
-      - name: test
-        env:
-          SLOW_TEST: true
-        run: |
-          make check
+  build:
+    uses: ./.github/workflows/build.yaml
```
rchincha, line 13: where is "slow-test" actually set to true?

smoser: > where is "slow-test" actually set to true?

It defaults to true. (.github/workflows/build.yaml).

it seems that "run all the tests" is a better default.  That way in a release and in a nightly we get all the tests run. the only place we don't get them all run is in the 'ci.yaml' caller.

smoser:
both of the slow test runs failed. see [here](https://github.com/project-stacker/stacker/actions/runs/6100746015/job/16555687757?pr=497).

unpriv failed:
```

smoser:
I'd like to ask that this be landed.  nightly c-i will still be broken, but this PR gets us closer.
Once landed, I think you'll be able to run the nightly manually on a branch to debug/fix it.
See https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow for how to do that.

PR #498, opened 2023-09-08
Doc/add some doc about make 

smoser:
*  Add some documentation to Makefile and doc/hacking.md.
    
    The docker.io bandwidth limit can easily stop a developer from being
    able to build or test docker.  Even without the limit, docker.io
    is much slower than a local zot or docker repository.
    
    Add documentation on how to set the makefile variable
    STACKER_DOCKER_BASE.
    
    Also add information about 'make test' and individual test files.

 *  Create a 'test' target separate from 'check'
    
    This just allows you to run 'make test'.  Previously the only
    way to run test from make was to call 'make check' which require
    running lint also.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,10 @@ PRIVILEGE_LEVEL?=
 # make check TEST=basic will run only the basic test
 # make check PRIVILEGE_LEVEL=unpriv will run only unprivileged tests
 .PHONY: check
-check: stacker lint $(REGCLIENT) $(ZOT)
+check: lint test
+
+.PHONY: test
+test: stacker $(REGCLIENT) $(ZOT)
 	sudo -E PATH="$$PATH" \
 		LXC_BRANCH=$(LXC_BRANCH) \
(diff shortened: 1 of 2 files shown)


smoser:
You can see the rendered doc/hacking.md https://github.com/smoser/stacker/blob/doc/add-some-doc/doc/hacking.md

smoser:
bump. 
I know there was a request in another channel that "All docs have to be ported over to the [stackerbuild.io](https://stackerbuild.io/) docs website".

a.) I'm not sure how to do that, please let me know how.
b.) i don't necessarily think that developer documentation (like HACKING.md) should live somewhere other than git
c.) if there is no magic sync of stuff in doc/ to stackerbuild.io, then we're really just making work for ourselves.  Having multiple sources of documentation maintained by stacker is just asking for inconsistencies.

My suggestion is:
a. keep developer documentation (doc related to modifying or building code) in git only
b. get rid of user doc in the tree or make make something automatically sync to stackerbuild.io so things aren't out of sync.

PR #509, opened 2023-09-21
Cleanup/imports parsing

smoser:
* fix: Improvements to yaml.Unmarshal for types.Imports and add some go tests
    
    Improvements here:
     * Drop duplicate code blocks from getImportFromInterface
       There were 2 identical codeblocks, one with a type convert to
       map[string]interface{} and one with type convert to
       map[interface{}]interface{}.  I cannot find any reason for the
       former
     * Add some golang tests to for unmarshalling.
     * Type check the values of all string Import fields, rather than just
       converting them to a string with 'fmt.Sprintf("%v")'
    

 * cleanup: Add contsants UidEmpty and GidEmpty and FileCopyNoPerms function.
    
    Instead of using '-1' to indicate "do not change userid" (or groupid)
    to FileCopy, provide constants UidEmpty and GidEmpty.
    
    Also add simplified FileCopyNoPerms to call FileCopy without mode
    or uid/gid.

### DIFF
--- a/pkg/lib/dir.go
+++ b/pkg/lib/dir.go
@@ -65,7 +65,7 @@ func DirCopy(dest string, source string) error {
 				return err
 			}
 		} else {
-			if err = FileCopy(dstfp, srcfp, nil, -1, -1); err != nil {
+			if err = FileCopyNoPerms(dstfp, srcfp); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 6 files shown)


rchincha (approved):
lgtm

rchincha:
@mikemccracken @hallyn pls review.

PR #511, opened 2023-09-22
feat: Better lint and test for devs and Makefile cleanup.

smoser:
This change does several things all towards the end of making
it easier for developers to run tests and lint on go code.

Now a developer can run either 'make go-test' or 'make lint'
without having lots of C dependencies or running in a specific
container or stacker.

Things changed:

 * Add a build tag 'skipembed' that builds without needing
   cmd/stacker/lxc-wrapper/lxc-wrapper . This allows you to
   build (or test) without that file.  In order to ensure that
   you do not get very far trying to run a 'stacker' binary,
   I've added a 'panic' if stacker binary is built with skipembed.

 * Add download-tools target to makefile to ease in downloading of
   regctl and zot.

 * Move the downloading of golangci-lint from the github workflow
   to Makefile.  This makes it easier for developer to get it.
   It is also added to the 'download-tools' target.

 * be specific about the version of golangci-lint that is used.
   1.54.2.  golangci-lint's docs specifically say to do this:

    > IMPORTANT: It's highly recommended installing a specific version
    > of golangci-lint available on the releases page.

   changing it is simply a matter of changing the value in the makefile.

 * Move running of 'go test' out of the lint target to its own
   'go-test' target.  go-test target will now write a coverage.html
   so you can view coverage easily.

 * add 'dlbin' makefile "function" for downloading binaries, and use
   that for regctl and zot from their rules.


rchamarthy (approved):

smoser:
I had to re-push to adjust a commit message comment. so af09c660aec829258d9ea5d402f6b0fb683208e7 is identical to the version @rchamarthy  reviewed ([ 934bce6 ](https://github.com/project-stacker/stacker/commit/934bce63f5fde474e66f6af672d90174cb1e0455)to af09c66) but has a different git message.

the difference in the commit message was just to drop the mention of moving coverage.txt as @rchincha asked that it not be moved.

PR #517, opened 2023-10-02
fix: Improve bind imports, support lower case 'source' in yaml, add go test

smoser:
Entries stacker.yaml's "binds" are of type Bind.
When imported from yaml, support was present for these three things:
1. ascii art specifying source and dest ("source -> dest")
2. source only ("source")
3. map with 'Source' and 'Dest'.

Support for '3' is dropped, and replaced by map with 'source' and 'dest'.  'Source' and 'Dest' are inconsistent with the rest of stacker.yaml. Note that importing from json still does support capital letter variants as the golang json library is case insensitive.

Improvements to the bind yaml and json support generally rely more heavily on the library to do the lifting.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/types/layer.go
+++ b/pkg/types/layer.go
@@ -2,7 +2,6 @@ package types
 
 import (
 	"encoding/json"
-	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


smoser:
builds failed, like below. I uploaded https://github.com/project-stacker/stacker/pull/518 and I'm hitting rebuids.

```
Couldn't obtain file info of https://github.com/lvmteam/lvm2/archive/refs/tags/v2_03_18.tar.gz, using cached copy
overlay-dirs, possibly modified after import: []
loading docker://alpine:edge
error: initializing source docker://alpine:edge: pinging container registry registry-1.docker.io: received unexpected HTTP status: 503 Service Unavailable
couldn't import base layer alpine 
```

pkg/types/layer.go
```diff
@@ -73,40 +72,6 @@ type Bom struct {
...
-
-	_, ok = bindSource.(string)
-	if !ok {
-		return nil, errors.Errorf("unknown bind source type, expected string: %T", i)
-	}
-
-	bindDest, ok := bindMap["Dest"]
-	if !ok {
-		return nil, errors.Errorf("bind dest missing: %v", i)
-	}
```
smoser, line 96: I guess previously we always required 'Dest' in the dictionary form.
We did *not* require it in the simple string form; it would simply make dest = source.   I find 'source' without 'dest' wierd, I did make it work though.  I'm willing to drop that function. While weird, it does allow you to "not repeat yourself" in the event that you wanted to bind a path inside same as outside.

hallyn (approved):

pkg/types/layer.go
smoser: well nothing depends on it now (or at least we did not have any tests prior to yesterday that tested it).
so we can drop it and not break any known behavior.

I'm feeling lazy, i guess just leave it in for now.  Point at this comment later when having done so causes problems to say "see, its your fault".

PR #519, opened 2023-10-03
use squashfuse ready notifier if available

hallyn:
When using squasfuse, check whether it supports -o notifypipe. If it does, then use that instead of our manual checking the mountpoint inode, which is less reliable.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


hallyn:
It's not the prettiest - I'd probably like to factor both the 'wait until ready' paths into separate functions.

hallyn:
closing in favor of #528

PR #521, opened 2023-10-04
feat(check): add a check for symlinks in roots-dir path

rchincha:
stacker tries to avoid symlinks to minimize unexpected behaviors, so add check for that.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/cmd/stacker/check.go
+++ b/cmd/stacker/check.go
@@ -4,6 +4,7 @@ import (
 	"io/fs"
 	"os"
 	"os/exec"
+	"path/filepath"
 
 	"github.com/pkg/errors"
 	"github.com/pkg/xattr"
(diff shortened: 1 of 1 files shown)


cmd/stacker/check.go
```diff
@@ -26,12 +27,23 @@ func doCheck(ctx *cli.Context) error {
...
 	}
 
+	// internally there are many checks to avoid symlinks
+	evalp, err := filepath.EvalSymlinks(config.RootFSDir)
+	if err != nil {
+		return errors.Wrapf(err, "%s: unable to evaluate path for symlinks", config.RootFSDir)
+	}
+
+	if evalp != config.RootFSDir {
+		return errors.Errorf("%s: roots dir (--roots-dir) path uses symbolic links, use %s instead", config.RootFSDir, evalp)
```
hallyn, line 43: Is it generally deemed safer to use %q for escaping paths?

rchincha: Fixed.

hallyn (approved):

PR #522, opened 2023-10-04
fix: handle ancient empty docker layers

rchincha:
Earlier versions of docker images had empty layers of 1024 zero-valued octets.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -46,6 +46,10 @@ func Infof(msg string, v ...interface{}) {
 	addStackerLogSentinel(log.NewEntry(log.Log.(*log.Logger))).Infof(msg, v...)
 }
 
+func Warnf(msg string, v ...interface{}) {
+	addStackerLogSentinel(log.NewEntry(log.Log.(*log.Logger))).Warnf(msg, v...)
+}
+
 func Errorf(msg string, v ...interface{}) {
 	addStackerLogSentinel(log.NewEntry(log.Log.(*log.Logger))).Errorf(msg, v...)
 }
--- a/pkg/overlay/metadata.go
+++ b/pkg/overlay/metadata.go
@@ -3,6 +3,7 @@ package overlay
 import (
 	"bytes"
 	"encoding/json"
+	"io/fs"
 	"os"
 	"path"
 
@@ -117,7 +118,15 @@ func (ovl overlayMetadata) lxcRootfsString(config types.StackerConfig, tag strin
 	for _, layer := range manifest.Layers {
 		contents := overlayPath(config.RootFSDir, layer.Digest, "overlay")
 		if _, err := os.Stat(contents); err != nil {
-			return "", errors.Wrapf(err, "%s does not exist", contents)
+			if errors.Is(err, fs.ErrNotExist) {
+				// some docker layers may be empty tars, so ignore these
+				// https://github.com/moby/moby/issues/20917#issuecomment-191901912
+				log.Warnf("%s skipping empty tar layer", layer.Digest)
+
+				continue
+			}
+
+			return "", errors.Wrapf(err, "%s unable to stat", contents)
 		}
 		lowerdirs = append(lowerdirs, contents)
 	}
--- a/pkg/stacker/base.go
+++ b/pkg/stacker/base.go
@@ -87,7 +87,11 @@ func SetupRootfs(o BaseLayerOpts) error {
 	case types.OCILayer:
 		fallthrough
 	case types.DockerLayer:
-		return setupContainersImageRootfs(o)
+		err := setupContainersImageRootfs(o)
+		if err != nil && errors.Is(err, types.ErrEmptyLayers) {
+			return o.Storage.SetupEmptyRootfs(o.Name)
+		}
+		return err
 	default:
 		return errors.Errorf("unknown layer type: %v", o.Layer.From.Type)
 	}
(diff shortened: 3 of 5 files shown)


hallyn:
@rchincha the only file in the only patch is a .bats file.  Did you forget some git-add action?

rchincha:
> @rchincha the only file in the only patch is a .bats file. Did you forget some git-add action?

3 files right?

test/docker-base.bats
```diff
@@ -31,3 +31,13 @@ EOF
     umoci unpack --image oci:layer1 dest
     [ ! -f dest/rootfs/favicon.ico ]
 }
+
+@test "image with empty layer" {
+    cat > stacker.yaml <<EOF
+image:
+    from:
+        type: docker
+        url: docker://grafana/grafana-oss:10.1.2-ubuntu
```
raharper, line 40: How do we know this image will stay around and will also have empty layers?  Would it be better to create an an image with an empty layer in the test-suite itself?

rchincha: Good question, we can push it to stacker ghcr to preserve it. Some strange combination of tooling - it is dockerv2 schema which still contains schemav1-type layers.

rchincha: https://github.com/project-stacker/stacker/pull/525

Would rather take an actual example.

rchincha: And not quite the same. The layer exists, but when you untar, it has zero changes to the rootfs.

raharper: OK, seems like we can create it somehow (docker certainly does).  

Right, so I copied out the empty tar.gz file from that graphana image...

Let me get that to you to test.

raharper: I think this should work, but you said it didn't.  Does stacker not detect the empty layer in here?

```
#!/bin/bash
OCID=oci_dir
umoci init --layout "$OCID"
umoci new --image "$OCID:emptylayer"

EMPTY_TGZ="empty-layer.tar.gz"
base64 --decode > "$EMPTY_TGZ" << EOF
H4sIAAAAAAAA/2IYBaNgFIxYAAgAAP//Lq+17wAEAAA=
EOF

umoci raw add-layer --image "$OCID:emptylayer" \
    --history.author docker \
    --history.comment "empty tar" \
    "$EMPTY_TGZ"
    
umoci gc --layout "$OCID"
```

maybe the GC helps?

rchincha: This seems to be creating an image with a single empty layer. Don't think there is any "legal" image that has that.

rchincha:
Without this PR, the same test fails.
https://github.com/project-stacker/stacker/pull/525

A real-world image has been copied over to ghcr.io so should be internal without any dependency on dockerhub.

raharper:
https://github.com/containers/image/blob/main/image/docker_schema2.go
https://github.com/containers/image/blob/main/internal/image/docker_schema2.go

I think this means it is a well-known type -- can we test if the image has the matching sha256? 

I'm going to see if I can programmatically create a simple image with such a layer.

rchincha:
> https://github.com/containers/image/blob/main/image/docker_schema2.go https://github.com/containers/image/blob/main/internal/image/docker_schema2.go
> 
> I think this means it is a well-known type -- can we test if the image has the matching sha256?
> 
> I'm going to see if I can programmatically create a simple image with such a layer.

A single layer doesn't make sense IMO. What Dockerfile directive would make only that?

rchincha:
https://github.com/containers/image/blob/main/internal/image/docker_schema2.go#L29

raharper:
> > https://github.com/containers/image/blob/main/image/docker_schema2.go https://github.com/containers/image/blob/main/internal/image/docker_schema2.go
> > I think this means it is a well-known type -- can we test if the image has the matching sha256?
> > I'm going to see if I can programmatically create a simple image with such a layer.
> 
> A single layer doesn't make sense IMO. What Dockerfile directive would make only that?

ENV or RUN; any command that doesn't actually modify the rootfs.  Why? I do not know.

rchincha:
> > > https://github.com/containers/image/blob/main/image/docker_schema2.go https://github.com/containers/image/blob/main/internal/image/docker_schema2.go
> > > I think this means it is a well-known type -- can we test if the image has the matching sha256?
> > > I'm going to see if I can programmatically create a simple image with such a layer.
> > 
> > 
> > A single layer doesn't make sense IMO. What Dockerfile directive would make only that?
> 
> ENV or RUN; any command that doesn't actually modify the rootfs. Why? I do not know.

We need at lea

rchincha:
```
FROM scratch
ENV a=b
```

gives us

```
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "config": {
    "digest": "sha256:2b4a446f5ab37b5895b62507c572303f16ab4f8d9161a35799c994301722d3d5",
    "size": 334
  },
  "layers": []
}
```

rchincha:
Included your example in the unit tests.

rchincha:
Will move the ghcr test to a SLOW_TEST (nightly) once you review/approve

raharper:
This PR should close #525 right?

rchincha:
> This PR should close #525 right?

Correct, PR #525 was only to demo that this PR is needed.

raharper (approved):

PR #528, opened 2023-10-19
feat: use squashfuse ready notifier if available

ariel-miculas:
When using squasfuse, check whether it supports -o notifypipe. If it does, then use that instead of our manual checking the mountpoint inode, which is less reliable.

Co-Developed-by: Serge Hallyn <serge@hallyn.com>

See [previous PR](https://github.com/project-stacker/stacker/pull/519)
Changes compared to the original PR:
* use a version check for figuring out whether the mount notification mechanism is supported
* check whether the pipe signals a successful or an unsuccessful squashfuse mount

This is a new PR because I cannot amend Serge's original PR.

**What type of PR is this?**
feature

none

It uses the newly available squashfuse notification mechanism.

Tested with the following program:
```
package main
import (
    "fmt"
    "os/exec"
    "strings"
    "path/filepath"
    "github.com/Masterminds/semver/v3"
    "syscall"
    "time"
    "os"
    "github.com/pkg/errors"
    )

func sqfuseSupportsMountNotification(sqfuse string) (bool) {
    fmt.Println("hello world")
    cmd := exec.Command("squashfuse")
    out, _ := cmd.CombinedOutput()
    // Don't care about errors
    first_line := strings.Split(string(out[:]), "\n")[0];
    version := strings.Split(first_line, " ")[1];
    v, err := semver.NewVersion(version)
    if err != nil {
        return false
    }
    // squashfuse notify mechanism was merged in 0.5.0
    constraint, err := semver.NewConstraint(">= 0.5.0");
    if err != nil {
        return false
    }
    if constraint.Check(v) {
        fmt.Printf("v: %s is at least than 0.5.0\n", version);
        return true
    }
    return false
}

func mountSqfs(squashFile string, mountpoint string) (*exec.Cmd, error) {
    sqfuse := "squashfuse"
    sqNotify := sqfuseSupportsMountNotification(sqfuse)
    var cmd *exec.Cmd
    var err error

    notifyOpts := ""
    notifyPath := ""
    if sqNotify {
        sockdir, err := os.MkdirTemp("", "sock")
        if err != nil {
            return cmd, err
        }
        notifyPath = filepath.Join(sockdir, "notifypipe")
        if err := syscall.Mkfifo(notifyPath, 0640); err != nil {
            return cmd, err
        }
        notifyOpts = "notify_pipe=" + notifyPath

    }

    optionArgs := "debug"
    if notifyOpts != "" {
        optionArgs += "," + notifyOpts
    }
    cmd = exec.Command(sqfuse, "-f", "-o", optionArgs, squashFile, mountpoint)
    cmd.Stdin = nil
    fmt.Printf("Extracting %s -> %s with %s\n", squashFile, mountpoint, sqfuse)
    err = cmd.Start()
    if err != nil {
        return cmd, err
    }

    // now poll/wait for one of 3 things to happen
    // a. child process exits - if it did, then some error has occurred.
    // b. the directory Entry is different than it was before the call
    //    to sqfuse.  We have to do this because we do not have another
    //    way to know when the mount has been populated.
    //    https://github.com/vasi/squashfuse/issues/49
    // c. a timeout (timeLimit) was hit
    timeLimit := 30 * time.Second
    alarmCh := make(chan struct{})
    go func() {
        cmd.Wait()
        close(alarmCh)
    }()
    if sqNotify {
        notifyCh := make(chan byte)
        fmt.Printf("%s supports notify pipe, watching %q\n", sqfuse, notifyPath)
        go func() {
            f, err := os.Open(notifyPath)
            if err != nil {
                return
            }
            defer f.Close()
            b1 := make([]byte, 1)
            for {
                n1, err := f.Read(b1)
                if err != nil {
                    return
                }
                if err == nil && n1 >= 1 {
                    fmt.Println(string(b1[:]))
                    break
                }
            }
            notifyCh <- b1[0]
        }()
        if err != nil {
            return cmd, errors.Wrapf(err, "Failed reading %q", notifyPath)
        }

        select {
        case <-alarmCh:
            fmt.Printf("Killing process %p\n", cmd.Process)
            cmd.Process.Kill()
            return cmd, errors.Wrapf(err, "Gave up on squashFuse mount of %s with %s after %s", squashFile, sqfuse, timeLimit)
        case ret := <-notifyCh:
            if ret == 's' {
                fmt.Println("success")
                return cmd, nil
            } else {
                fmt.Println("failure")
                return cmd, errors.New("squashfuse returned an error")
            }
        }
    }
    fmt.Printf("%s does not support notify pipe\n", sqfuse)

    return cmd, nil
}

func main() {
    if supported := sqfuseSupportsMountNotification("squashfuse"); supported {
        fmt.Println("squashfuse supports notification!");
    } else {
        fmt.Println("squashfuse doesn't support notification!");
    }

    squashFile := "/home/amiculas/work/cisco/test-puzzlefs/barehost.sqhs"
    mountpoint := "/tmp/sqfs-mount"
    _, err := mountSqfs(squashFile, mountpoint);
    if err != nil {
        fmt.Println("error detected:", err)
        return
    }
}
```
Output:
```
$ go run mount.go
hello world
v: 0.5.0 is at least than 0.5.0
squashfuse supports notification!
hello world
v: 0.5.0 is at least than 0.5.0
Extracting /home/amiculas/work/cisco/test-puzzlefs/barehost.sqhs -> /tmp/sqfs-mount with squashfuse
squashfuse supports notify pipe, watching "/tmp/sock2231501866/notifypipe"
s
success
~/work/hello-go/squash-mount

$ go run mount.go
hello world
v: 0.5.0 is at least than 0.5.0
squashfuse supports notification!
hello world
v: 0.5.0 is at least than 0.5.0
Extracting /home/amiculas/work/cisco/test-puzzlefs/barehost.sqhs -> /tmp/sqfs-mount with squashfuse
squashfuse supports notify pipe, watching "/tmp/sock2289448691/notifypipe"
f
failure
error detected: squashfuse returned an error
```
The first time it succeeds and the second time it fails, since the fuse mountpoint already exists.
## With stacker:
```
amiculas@ubuntu-vm:~/stacker$ cat hello-stacker
hello-stacker:
  from:
    type: docker
    url: docker://zothub.io/tools/busybox:stable
  run: |
    mkdir -p /hello-stacker-app
    echo 'echo "Hello Stacker!"' > /hello-stacker-app/hello.sh
    chmod +x /hello-stacker-app/hello.sh
  entrypoint: /hello-stacker-app/hello.sh

amiculas@ubuntu-vm:~/stacker$ ./stacker build --layer-type=squashfs -f hello-stacker
```
### Old squashfuse (<0.5.0)
```
amiculas@ubuntu-vm:~/stacker$ which squashfuse
/usr/bin/squashfuse

amiculas@ubuntu-vm:~/stacker$ squashfuse
squashfuse 0.1.103 (c) 2012 Dave Vasilevsky

amiculas@ubuntu-vm:~/stacker$ ./stacker internal-go atomfs mount hello-stacker-squashfs sqfs
/usr/bin/squashfuse does not support notify pipe
/usr/bin/squashfuse does not support notify pipe
error: couldn't do overlay mount to sqfs, opts: index=off,xino=on,userxattr,lowerdir=/home/amiculas/stacker/atomfs-metadata/mounts/8c4a55c9fd4dbc29a4f9d47d663a73753c877ef4419835f90afb497b8e9907c1:/home/amiculas/stacker/atomfs-metadata/mounts/f20ee95456ef2f3c9761a8e10f1279331980bd6d53cac785294be19ff6fccb55: operation not permitted
```
Note that the overlay fails, but the squashfuse mounts have been created:
```
amiculas@ubuntu-vm:~/stacker$ mount | grep fuse.squashfuse
squashfuse on /home/amiculas/stacker/atomfs-metadata/mounts/8c4a55c9fd4dbc29a4f9d47d663a73753c877ef4419835f90afb497b8e9907c1 type fuse.squashfuse (rw,nosuid,nodev,relatime,user_id=1001,group_id=1001,allow_other)
squashfuse on /home/amiculas/stacker/atomfs-metadata/mounts/f20ee95456ef2f3c9761a8e10f1279331980bd6d53cac785294be19ff6fccb55 type fuse.squashfuse (rw,nosuid,nodev,relatime,user_id=1001,group_id=1001,allow_other)
```
### New squashfuse (>=0.5.0)
```
amiculas@ubuntu-vm:~/stacker$ which squashfuse
/usr/local/bin/squashfuse

amiculas@ubuntu-vm:~/stacker$ squashfuse
squashfuse 0.5.0 (c) 2012 Dave Vasilevsky

amiculas@ubuntu-vm:~/stacker$ which squashfuse_ll
/usr/local/bin/squashfuse_ll

amiculas@ubuntu-vm:~/stacker$ squashfuse_ll
squashfuse 0.5.0 (c) 2012 Dave Vasilevsky

amiculas@ubuntu-vm:~/stacker$ ./stacker internal-go atomfs mount hello-stacker-squashfs sqfs
/usr/local/bin/squashfuse_ll supports notify pipe, watching "/tmp/sock3579536881/notifypipe"
/usr/local/bin/squashfuse_ll supports notify pipe, watching "/tmp/sock4245449089/notifypipe"
error: couldn't do overlay mount to sqfs, opts: index=off,xino=on,userxattr,lowerdir=/home/amiculas/stacker/atomfs-metadata/mounts/8c4a55c9fd4dbc29a4f9d47d663a73753c877ef4419835f90afb497b8e9907c1:/home/amiculas/stacker/atomfs-metadata/mounts/f20ee95456ef2f3c9761a8e10f1279331980bd6d53cac785294be19ff6fccb55: operation not permitted
```
Note that the overlay fails, but the squashfuse mounts have been created:
```
amiculas@ubuntu-vm:~/stacker$ mount | grep fuse.squashfuse
squashfuse_ll on /home/amiculas/stacker/atomfs-metadata/mounts/8c4a55c9fd4dbc29a4f9d47d663a73753c877ef4419835f90afb497b8e9907c1 type fuse.squashfuse_ll (rw,nosuid,nodev,relatime,user_id=1001,group_id=1001,allow_other)
squashfuse_ll on /home/amiculas/stacker/atomfs-metadata/mounts/f20ee95456ef2f3c9761a8e10f1279331980bd6d53cac785294be19ff6fccb55 type fuse.squashfuse_ll (rw,nosuid,nodev,relatime,user_id=1001,group_id=1001,allow_other)
```

**Automation added to e2e**:

No

No

### DIFF
--- a/pkg/squashfs/squashfs.go
+++ b/pkg/squashfs/squashfs.go
@@ -15,6 +15,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
 	"stackerbuild.io/stacker/pkg/log"
@@ -42,6 +43,15 @@ type ExcludePaths struct {
 	exclude map[string]bool
 	include []string
 }
+
+type squashFuseInfoStruct struct {
+	Path           string
+	Version        string
+	SupportsNotfiy bool
+}
+
+var once sync.Once
+var squashFuseInfo = squashFuseInfoStruct{"", "", false}
 
 func NewExcludePaths() *ExcludePaths {
 	return &ExcludePaths{
@@ -198,11 +208,45 @@ func isMountedAtDir(src, dest string) (bool, error) {
 	return false, nil
 }
 
-func findSquashfusePath() string {
+func findSquashFuseInfo() {
+	var sqfsPath string
 	if p := which("squashfuse_ll"); p != "" {
-		return p
+		sqfsPath = p
+	} else {
+		sqfsPath = which("squashfuse")
 	}
-	return which("squashfuse")
+	if sqfsPath == "" {
+		return
+	}
+	version, supportsNotify := sqfuseSupportsMountNotification(sqfsPath)
+	log.Infof("Found squashfuse at %s (version=%s notify=%t)", sqfsPath, version, supportsNotify)
+	squashFuseInfo = squashFuseInfoStruct{sqfsPath, version, supportsNotify}
+}
+
+// sqfuseSupportsMountNotification - returns true if squashfuse supports mount
+// notification, false otherwise
+// sqfuse is the path to the squashfuse binary
+func sqfuseSupportsMountNotification(sqfuse string) (string, bool) {
+	cmd := exec.Command(sqfuse)
+
+	// `squashfuse` always returns an error...  so we ignore it.
+	out, _ := cmd.CombinedOutput()
+
+	firstLine := strings.Split(string(out[:]), "\n")[0]
+	version := strings.Split(firstLine, " ")[1]
+	v, err := semver.NewVersion(version)
+	if err != nil {
+		return version, false
+	}
+	// squashfuse notify mechanism was merged in 0.5.0
+	constraint, err := semver.NewConstraint(">= 0.5.0")
+	if err != nil {
+		return version, false
+	}
+	if constraint.Check(v) {
+		return version, true
+	}
+	return version, false
 }
 
 var squashNotFound = errors.Errorf("squashfuse program not found")
@@ -211,10 +255,26 @@ var squashNotFound = errors.Errorf("squashfuse program not found")
 // return a pointer to the squashfuse cmd.
 // The caller of the this is responsible for the process created.
 func squashFuse(squashFile, extractDir string) (*exec.Cmd, error) {
-	sqfuse := findSquashfusePath()
 	var cmd *exec.Cmd
-	if sqfuse == "" {
+
+	once.Do(findSquashFuseInfo)
+	if squashFuseInfo.Path == "" {
 		return cmd, squashNotFound
+	}
+
+	notifyOpts := ""
+	notifyPath := ""
+	if squashFuseInfo.SupportsNotfiy {
+		sockdir, err := os.MkdirTemp("", "sock")
+		if err != nil {
+			return cmd, err
+		}
+		defer os.RemoveAll(sockdir)
+		notifyPath = filepath.Join(sockdir, "notifypipe")
+		if err := syscall.Mkfifo(notifyPath, 0640); err != nil {
+			return cmd, err
+		}
+		notifyOpts = "notify_pipe=" + notifyPath
 	}
 
 	// given extractDir of path/to/some/dir[/], log to path/to/some/.dir-squashfs.log
@@ -240,12 +300,16 @@ func squashFuse(squashFile, extractDir string) (*exec.Cmd, error) {
 	// It would be nice to only enable debug (or maybe to only log to file at all)
 	// if 'stacker --debug', but we do not have access to that info here.
 	// to debug squashfuse, use "allow_other,debug"
-	cmd = exec.Command(sqfuse, "-f", "-o", "allow_other,debug", squashFile, extractDir)
+	optionArgs := "allow_other,debug"
+	if notifyOpts != "" {
+		optionArgs += "," + notifyOpts
+	}
+	cmd = exec.Command(squashFuseInfo.Path, "-f", "-o", optionArgs, squashFile, extractDir)
 	cmd.Stdin = nil
 	cmd.Stdout = cmdOut
 	cmd.Stderr = cmdOut
 	cmdOut.Write([]byte(fmt.Sprintf("# %s\n", strings.Join(cmd.Args, " "))))
-	log.Debugf("Extracting %s -> %s with %s [%s]", squashFile, extractDir, sqfuse, logf)
+	log.Debugf("Extracting %s -> %s with %s [%s]", squashFile, extractDir, squashFuseInfo.Path, logf)
 	err = cmd.Start()
 	if err != nil {
 		return cmd, err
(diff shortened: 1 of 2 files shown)


smoser:
> If it does, then use that instead of our manual checking the mountpoint inode, which is less reliable.

Is it actually less reliable? I agree that responding to an event is nicer than polling, but have we actually found polling to be "unreliable"?

smoser (commented):
I think this looks sane.
It'd be nice if we could only check once (sync.Once) for presense of squashfuse executable and also whether or not it has notify pipe support.

I don't think its a blocker (we're currently checking squashfuse every time) but would be nice.

ariel-miculas:
> > If it does, then use that instead of our manual checking the mountpoint inode, which is less reliable.
> 
> Is it actually less reliable? I agree that responding to an event is nicer than polling, but have we actually found polling to be "unreliable"?

These were Serge's words, @hallyn do you have any comments?

ariel-miculas:
> I think this looks sane. It'd be nice if we could only check once (sync.Once) for presense of squashfuse executable and also whether or not it has notify pipe support.
> 
> I don't think its a blocker (we're currently checking squashfuse every time) but would be nice.

Well, `findSquashfusePath` doesn't  execute the squashfuse binary, but I guess it could return a tuple of (Path, Version) instead of only the Path.

smoser:
> Well, findSquashfusePath doesn't execute the squashfuse binary, but I guess it could return a tuple of (Path, Version) instead of only the Path.

yeah, i'd be fine with that. and it can just store the result in a global

```golang
type squashfuse struct {
   Path string
   Version version
   notfiy bool
}
```

oir something like that.

ariel-miculas:
> > Well, findSquashfusePath doesn't execute the squashfuse binary, but I guess it could return a tuple of (Path, Version) instead of only the Path.
> 
> yeah, i'd be fine with that. and it can just store the result in a global
> 
> ```go
> type squashfuse struct {
>    Path string
>    Version version
>    notfiy bool
> }
> ```
> 
> oir something like that.

And initialize this struct only once? Then, for subsequent calls to the findSquashfusePath function, just return the global struct?

hallyn:
> > If it does, then use that instead of our manual checking the mountpoint inode, which is less reliable.
> 
> Is it actually less reliable? I agree that responding to an event is nicer than polling, but have we actually found polling to be "unreliable"?

Yes.  Because it checks inode version of the root dentry, I have seen cases where, when already inside an overlayfs, the mount hung because the dentry info never appeared to change.

At least, I think that's what I saw :)  Writing a reliable reproducer of that would be kind of nifty.

hallyn:
> > > Well, findSquashfusePath doesn't execute the squashfuse binary, but I guess it could return a tuple of (Path, Version) instead of only the Path.
> > 
> > 
> > yeah, i'd be fine with that. and it can just store the result in a global
> > ```
> > type squashfuse struct {
> >    Path string
> >    Version version
> >    notfiy bool
> > }
> > ```
> > 
> > 
> > oir something like that.
> 
> And initialize this struct only once? Then, for subsequent calls to the findSquashfusePath function, just return the global struct?

Right, if you initialize it in a [sync.Once](https://pkg.go.dev/sync#Once), it'll automatically just happen once.

hallyn:
@smoser @ariel-miculas it appears that every overlay mount's root dentry shows a different 'device' in stat(), and that's enough for os.SameFile() to do the right thing.

So the only advantage here is not waiting (not waiting too long, and not risking timeing out too soon).  I still think that's worthwhile.  But there may not be a correctness issue as I thought there was.

hallyn:
> @hallyn what does overlay have to do with squashfuse?

Right you are, I was confusing myself.

pkg/squashfs/squashfs.go
```diff
@@ -211,12 +254,27 @@ var squashNotFound = errors.Errorf("squashfuse program not found")
...
+
+	once.Do(findSquashFuseInfo)
+	if squashFuseInfo.Path == "" {
 		return cmd, squashNotFound
 	}
 
+	notifyOpts := ""
+	notifyPath := ""
+	if squashFuseInfo.SupportsNotfiy {
+		sockdir, err := os.MkdirTemp("", "sock")
```
hallyn, line 267: What's going to clean this directory up?

Can we do a "defer os.RemoveAll(sockdir)"?  After this function returns, the socket is no longer needed, right?

ariel-miculas: Sorry, I'm not used to cleaning up resourcing manually. I will add the defer.

pkg/squashfs/squashfs.go
```diff
@@ -211,12 +254,28 @@ var squashNotFound = errors.Errorf("squashfuse program not found")
...
+		defer os.RemoveAll(sockdir)
+		notifyPath = filepath.Join(sockdir, "notifypipe")
+		if err := syscall.Mkfifo(notifyPath, 0640); err != nil {
+			return cmd, err
+		}
+		notifyOpts = "notify_pipe=" + notifyPath
+	}
+
 	// given extractDir of path/to/some/dir[/], log to path/to/some/.dir-squashfs.log
 	extractDir = strings.TrimSuffix(extractDir, "/")
```
smoser, line 280: you could put the socket alongside the squashfs log as path/to/some/.dir-squashfuse.socket (and rename the '.dir-squashfs.log' to better '.dir-squash*fuse*.log' at the same time.

ariel-miculas: Well the log is there so it's persistent and you could see what the errors were, I don't think we need persistance for the socket.

pkg/squashfs/squashfs.go
```diff
@@ -198,11 +208,44 @@ func isMountedAtDir(src, dest string) (bool, error) {
...
+		sqfsPath = p
+	} else {
+		sqfsPath = which("squashfuse")
 	}
-	return which("squashfuse")
+	if sqfsPath == "" {
+		return
+	}
+	version, supportsNotfiy := sqfuseSupportsMountNotification(sqfsPath)
+	squashFuseInfo = squashFuseInfoStruct{sqfsPath, version, supportsNotfiy}
```
smoser, line 222: I think it would be good to have a Info level log here that says something like:

log.Infof("Found squashfuse at %s (version=%s notify=%t)", squashFuseInfo.Path, squashFuseInfo.version, squashFuseInfo.supportsNotify)

pkg/squashfs/squashfs.go
hallyn: "How barbaric!"

https://www.youtube.com/watch?v=LkqiDu1BQXY

hallyn (approved):

ariel-miculas:
@smoser, does the PR look good to you now?

PR #529, opened 2023-10-19
fix(gzip)!: change the default block size

rchincha:
BREAKING CHANGE: the default gzip block size is changed to 256<<12, was previously 256<<10.

The layers created by ecosystem tools and stacker are different for the same tar content. Unfortunately, the OCI image spec doesn't standardize/encode this in the document. Hence, we standardize in the implementation.

We now link against our own fork: github.com/project-stacker/umoci which may change depending on the PR getting merged to upstream.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


pkg/overlay/pack.go
```diff
@@ -408,7 +408,7 @@ func generateLayer(config types.StackerConfig, oci casext.Engine, mutators []*mu
 		defer blob.Close()
 
 		if layerType.Type == "tar" {
-			desc, err = mutator.Add(context.Background(), mediaType, blob, history, mutate.GzipCompressor, nil)
+			desc, err = mutator.Add(context.Background(), mediaType, blob, history, mutate.GzipCompressor.WithOpt(mutate.GzipBlockSize(256<<12)), nil)
```
raharper, line 411: Could we move the size to a const somewhere that's searchable, also with a comment indicating what it was previously and why we're changing it now?

rchincha: The commit msg/log captures that. But I can add a note.
Fixing the const.

rchincha: Fixed.

raharper (commented):
Suggestion in commit message:

A tar layer with the same content but compressed with different gzip blocksize
will result in different sha256sums in the final OCI Image.  Ecosystem tools
have one current size in use and stacker's current size differ.

Interactions between a stacker-built OCI image and ecosystem tools
which *recompress* lower layers results in bloated registries which will have
identical tar content but different compressed sha256 blobs.

Unfortunately, the OCI image spec doesn't standardize/encode this in the
specification document. Hence, we change to the current common block size used
in the ecosystem here in the stacker implementation.

rchincha:
> Suggestion in commit message:

Updated.

raharper (approved):

PR #531, opened 2023-10-25
fix(umoci)!: umoci's copy buffer set to 1 MiB

rchincha:
https://github.com/project-stacker/umoci/commit/89a7ea1a89bd405f527ea60a8c8ceaa38b412084

Use updated umoci (project-stacker's fork) to use a 1 MiB copy buffer. This change brings stacker inline with compression parameters used in containers/image libraries and makes them compatible.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,8 @@ HACK_D := $(TOP_LEVEL)/hack
 TOOLS_D := $(HACK_D)/tools
 REGCLIENT := $(TOOLS_D)/bin/regctl
 REGCLIENT_VERSION := v0.5.1
+export SKOPEO = $(TOOLS_D)/bin/skopeo
+export SKOPEO_VERSION = 1.9.3
 # OCI registry
 ZOT := $(TOOLS_D)/bin/zot
 ZOT_VERSION := v2.0.0-rc6
@@ -99,6 +101,19 @@ $(REGCLIENT):
 $(ZOT):
 	$(call dlbin,$@,https://github.com/project-zot/zot/releases/download/$(ZOT_VERSION)/zot-linux-amd64-minimal)
 
+$(SKOPEO):
+	@mkdir -p "$(TOOLS_D)/bin"; \
+	tmpdir=$$(mktemp -d); \
+	cd $$tmpdir; \
+	git clone https://github.com/containers/skopeo.git; \
+	cd skopeo; \
+	git fetch --all --tags --prune; \
+	git checkout tags/v$(SKOPEO_VERSION) -b tag-$(SKOPEO_VERSION); \
+	make bin/skopeo; \
+	cp bin/skopeo $(SKOPEO); \
+	cd $(TOP_LEVEL); \
+	rm -rf $$tmpdir;
+
 TEST?=$(patsubst test/%.bats,%,$(wildcard test/*.bats))
 PRIVILEGE_LEVEL?=
 
@@ -108,7 +123,7 @@ PRIVILEGE_LEVEL?=
 check: lint test go-test
 
 .PHONY: test
-test: stacker $(REGCLIENT) $(ZOT)
+test: stacker $(REGCLIENT) $(SKOPEO) $(ZOT)
 	sudo -E PATH="$$PATH" \
 		LXC_BRANCH=$(LXC_BRANCH) \
 		LXC_CLONE_URL=$(LXC_CLONE_URL) \
--- a/install-build-deps.sh
+++ b/install-build-deps.sh
@@ -3,12 +3,18 @@ set -o pipefail
 set -o errexit
 
 installdeps_fedora() {
-        sudo dnf install \
-        bats \
-        jq \
-        lxc-devel \
-        libcap-devel \
-        libacl-devel
+  sudo dnf install \
+    bats \
+    jq \
+    lxc-devel \
+    libcap-devel \
+    libacl-devel
+  # skopeo deps
+  sudo dnf install \
+    gpgme-devel \
+    libassuan-devel \
+    btrfs-progs-devel \
+    device-mapper-devel
     if ! command -v go 2>/dev/null; then
         sudo dnf install golang
         go version
@@ -35,9 +41,15 @@ installdeps_ubuntu() {
             lxc-utils \
             parallel \
             pkg-config \
-            skopeo \
             squashfs-tools \
             squashfuse
+    # skopeo deps
+    sudo apt -yy install \
+       libgpgme-dev \
+       libassuan-dev \
+       libbtrfs-dev \
+       libdevmapper-dev \
+       pkg-config
     if ! command -v go 2>/dev/null; then
         sudo apt -yy install golang-go
         go version
--- a/test/gzip.bats
+++ b/test/gzip.bats
@@ -1,0 +1,16 @@
+load helpers
+
+function setup() {
+    stacker_setup
+}
+
+function teardown() {
+    cleanup
+}
+
+@test "import various sizes" {
+  test_copy_buffer_size 512k tar
+  test_copy_buffer_size 2m tar
+  test_copy_buffer_size 512k tar.gz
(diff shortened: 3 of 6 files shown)


rchamarthy (changes requested):
Is this generically applicable to stacker world or is it a specific mixed case we are trying to solve?

rchincha:
> Is this generically applicable to stacker world or is it a specific mixed case we are trying to solve?

Targeting ecosystem compatibility.

https://github.com/opencontainers/image-spec/issues/1145
^ generic solution will need this to be fixed.

test/gzip.bats
```diff
@@ -0,0 +1,29 @@
...
+  cat > stacker.yaml <<EOF
+first:
+  from:
+    type: tar
+    url: test.tar
+EOF
+  stacker build
+  m1=$(cat oci/index.json | jq .manifests[0].digest | sed  's/sha256://' | tr -d \")
+  l1=$(cat oci/blobs/sha256/"$m1" | jq .layers[0].digest | sed  's/sha256://' | tr -d \")
+  skopeo copy --format=oci oci:oci:first containers-storage:test:tar
```
raharper, line 24: Does the version of skopeo matter (since it also uses the image golibrary) ?

Makefile
```diff
@@ -31,6 +31,8 @@ HACK_D := $(TOP_LEVEL)/hack
 TOOLS_D := $(HACK_D)/tools
 REGCLIENT := $(TOOLS_D)/bin/regctl
 REGCLIENT_VERSION := v0.5.1
+SKOPEO := $(TOOLS_D)/bin/skopeo
+SKOPEO_VERSION := v1.9.1
```
raharper, line 35:
```suggestion
export SKOPEO_VERSION = v1.9.1
```

test/gzip.bats
```diff
@@ -0,0 +1,36 @@
...
+  from:
+    type: tar
+    url: test.tar
+EOF
+  stacker build
+  cat oci/index.json | jq .
+  m1=$(cat oci/index.json | jq .manifests[0].digest | sed  's/sha256://' | tr -d \")
+  cat oci/blobs/sha256/"$m1" | jq .
+  l1=$(cat oci/blobs/sha256/"$m1" | jq .layers[0].digest | sed  's/sha256://' | tr -d \")
+  skopeo --version
```
raharper, line 26:
```suggestion
 $SKOPEO --version
 [[ "$(SKOPEO --version)" =~ "skopeo version ${SKOPEO_VERSION}" ] || {
     echo "$SKOPEO --version should be ${SKOPEO_VERSION}"
     exit 1
}
```

test/gzip.bats
```diff
@@ -0,0 +1,36 @@
...
+  skopeo copy --format=oci oci:oci:tar containers-storage:test:tar
+  skopeo copy --format=oci containers-storage:test:tar oci:oci:test
```
raharper, line 28:
```suggestion
  $SKOPEO copy --format=oci oci:oci:tar containers-storage:test:tar
  $SKOPEO copy --format=oci containers-storage:test:tar oci:oci:test
```

test/gzip.bats
rchincha: Fixed and added more tests.

install-build-deps.sh
```diff
@@ -38,6 +44,13 @@ installdeps_ubuntu() {
             skopeo \
```
raharper, line 44: You can drop skopeo here since we're building it now

test/gzip.bats
```diff
@@ -0,0 +1,129 @@
...
+  $SKOPEO copy --format=oci oci:oci:tar containers-storage:test:tar
+  $SKOPEO copy --format=oci containers-storage:test:tar oci:oci:test
+  cat oci/index.json | jq .
+  m2=$(cat oci/index.json | jq .manifests[1].digest | sed  's/sha256://' | tr -d \")
+  cat oci/blobs/sha256/"$m2" | jq .
+  l2=$(cat oci/blobs/sha256/"$m2" | jq .layers[0].digest | sed  's/sha256://' | tr -d \")
+  echo "$l1"
+  echo "$l2"
+  [ "$l1" = "$l2" ]
+}
```
raharper, line 39: Can we make a single function which takes the test params?

```
test_copy_buffer_size()
   local buffer_size = "$1"
   rm -rf "folder1"
   mkdir folder1
   truncate -s "$buffer_size" folder1/file1
   tar cvf test.tar folder1
   cat > stacker.yaml <<EOF
   ....
   [ "$l1" = "$l2" ]
}
```

Then do

```
@test "import tar > 1M" test_copy_buffer_size "2M"
```

or incase the `@test` doesn't accept params we can

```
test_copy_buffer_2M() {
    test_copy_buffer_size "2M"
}

@test "import tar > 1M" test_copy_buffer_2M
```

This makes it easier to see where the variablity lies and makes adjusting
the test case simpler (fix in one spot)

test/gzip.bats
rchincha: Fixed.

install-build-deps.sh
```diff
@@ -38,6 +44,13 @@ installdeps_ubuntu() {
...
             squashfs-tools \
             squashfuse
+    # skopeo deps
+    sudo apt -yy install \
```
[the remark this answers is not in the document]
rchincha: Wanted to keep the list separate so that if we decide to drop/update skopeo in the future, it will be that one line to fix.

test/gzip.bats
```diff
@@ -0,0 +1,129 @@
...
+
+@test "import tar.gz < 1M" {
+  mkdir -p folder1
+  truncate -s 512k folder1/file1
+  tar cvzf test.tar.gz folder1
+  cat > stacker.yaml <<EOF
```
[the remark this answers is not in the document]
rchincha: Independent tests - so we test for various sizes.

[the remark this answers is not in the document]
rchincha: and .tar and .tar.gz just to be sure.

test/gzip.bats
rchincha: Let me check this

test/gzip.bats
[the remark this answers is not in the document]
raharper: OK.  I'd try to use a tmpdir so that we know we're not mixing contents between tests.
w.r.t the sizes, there are no issues (AFAIK via my pgzip testing) with files (no tar) smaller than the readbuffer size (1M in our case now).  For tar.gz to be useful (I don't think it is, it's just like tar though less likely to compress any more) your tar.gz needs to be enough bigger than readBuffer size to ensure that we end up reading  two or more blocks (ie, pgzip will actually split up the input and compress in parallel).

test/helpers.bash
```diff
@@ -200,3 +200,37 @@ function zot_teardown {
...
+   local file_type="$2"
+   rm -rf folder1
+   mkdir folder1
+   truncate -s "$buffer_size" folder1/file1
+   tar cvf test."$file_type" folder1
+   cat > stacker.yaml <<EOF
+tar:
+  from:
+    type: tar
+    url: test."$file_type"
```
raharper, line 215:
```suggestion
    url: test.$file_type
```

This will fix it, but you might want to use stacker --substitute =)

test/gzip.bats
rchincha: It could, so just making sure we have a sane baseline.

test/gzip.bats
rchincha: Fixed.

rchincha:
$BATS_TEST_TMPDIR is a temporary directory unique for each test. Could be used to create files required only for specific tests.

test/gzip.bats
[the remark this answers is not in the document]
rchincha: Fixed.

raharper (approved):
Lgtm

PR #536, opened 2023-11-08
fix(umoci)!: umoci's copy buffer set to 1 MiB (#531)

rchincha:
* fix(umoci)!: umoci's copy buffer set to 1 MiB

https://github.com/project-stacker/umoci/commit/89a7ea1a89bd405f527ea60a8c8ceaa38b412084

Use updated umoci (project-stacker's fork) to use a 1 MiB copy buffer. This change brings stacker inline with compression parameters used in containers/image libraries and makes them compatible.

* test(gzip): add unit tests for gzip

---------

(cherry picked from commit 589a6486aeb2a5fc157edf043d0310795cd2c90b)

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


smoser:
I had just done https://github.com/project-stacker/stacker/pull/537 (and then saw that you did this).

I'm surprised that your branch could work.  the test you added I do not think can pass because it tests that it is using SKOPEO at a certain version, and you're not going to build that version.

i think mine should get skopeo built.

hallyn:
> I'm surprised that your branch could work

Hm, he didn't add SKPOEO to the make targets so maybe it didn't actually build it?

hallyn:
Do you need to add the build-deps as in Scott's pr?

PR #543, opened 2023-11-11
feat: Support legacy 'import' directive and use 'imports' for future

smoser:
This changes the behavior of the 'import' target to behave as it did before the breaking change that moved imports into /stacker/imports/

So now, if the stacker file uses 'import', then imports will be placed in /stacker.  If the stacker file uses 'imports' (plural) then they will be placed in /stacker/imports.

What we actually get in both cases is all the binds being done into either /.stacker (legacy) or /stacker (new).  In the legacy case, the imports are also bind-mounted into /stacker

Legacy (for those that used 'import:')

     stacker: /.stacker/bin/stacker
     imports: /.stacker/imports -> /stacker
     runscript: /.stacker/imports/.stacker-run.sh
     artifacts: /.stacker/artifacts

new (for those that use 'imports:')

     stacker: /stacker/bin/stacker
     imports: /stacker/imports
     runscript: /stacker/imports/.stacker-run.sh
     artifacts: /stacker/artifacts

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -3,7 +3,7 @@ build-env:
   from:
     type: docker
     url: ${{STACKER_BUILD_BASE_IMAGE}}
-  import:
+  imports:
     - https://github.com/json-c/json-c/archive/refs/tags/json-c-0.16-20220414.tar.gz
     - https://gitlab.com/cryptsetup/cryptsetup/-/archive/v2.6.0/cryptsetup-v2.6.0.tar.gz
     - https://github.com/lvmteam/lvm2/archive/refs/tags/v2_03_18.tar.gz
--- a/cmd/stacker/bom.go
+++ b/cmd/stacker/bom.go
@@ -10,6 +10,7 @@ import (
 	"stackerbuild.io/stacker-bom/pkg/bom"
 	"stackerbuild.io/stacker-bom/pkg/distro"
 	"stackerbuild.io/stacker-bom/pkg/fs"
+	"stackerbuild.io/stacker/pkg/types"
 )
 
 var bomCmd = cli.Command{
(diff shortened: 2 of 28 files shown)


smoser:
This wont work right now, I didn't adjust any of the tests, but this is the 'import -> /stacker' and 'imports -> /stacker/imports' change that I spoke with people about this week.

@mikemccracken , @hallyn , @rchincha

rchincha:
Would also add a "Warning" that this is to be deprecated.

smoser:
> Would also add a "Warning" that this is to be deprecated.

done.

smoser:
@rchincha  or @hallyn I'd appreciate feedback here.

This seems like a win to me.  Basically 
1. gets us backwards compatibility with "old stacker" (0.4.x) branch 
2. makes stacker files more consistent.  the following directives in a layer are all plural: overlay_dirs, volumes, labels, generate_labels, binds, annotations.  import was singular, the change makes it plural going forward.
3. Allows users to move to newer stacker and move to new 'imports' lazily.  the moves can even be done on one layer at a time (inside the same stacker file you can use 'import' for one layer and 'imports' for another).
4. cleans up the "binds" code a bit.


Note, it *does* break things that had moved to 1.0.0-rc5+ (where 'import' imported into /stacker/imports).  Consumers moving from 1.0.0-rc5+ will have to change 'import:' to 'imports:' in stacker files.  Consumers moving from < 1.0.0-rc5 should not need any changes.

Also note, we get a warning to console about deprecation of the 'import'.  I would personally like to change that warning to put a stick in the mud saying "No stacker release after 2024-01-01 will support 'import'".  That gives anyone picking this up a year to change and after that we can remove the old support.

rchincha (approved):
lgtm

PR #546, opened 2023-11-13
ci: add a job to sync 3rd party container images

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/.github/workflows/sync-3rdparty-images.yaml
+++ b/.github/workflows/sync-3rdparty-images.yaml
@@ -1,0 +1,35 @@
+name: 'Sync images and artifacts to ghcr'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+permissions: read-all
+
+jobs:
+  sync:
+    name: "images"
+    permissions:
+      contents: read
+      packages: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Log in to GitHub Docker Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Tag and push to ghcr
+        run: |
+          set -e
+          for n in ubuntu:latest alpine:edge centos:latest busybox:latest; do
+            dest="${{ github.repository_owner }}/$n"
+            docker trust inspect "$n"
+            docker pull $n
+            docker tag $n "$dest"
+            docker push $dest
+          done


.github/workflows/sync-3rdparty-images.yaml
```diff
@@ -0,0 +1,104 @@
...
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+permissions: read-all
+
+jobs:
+  sync-golang:
+    name: 'golang'
```
smoser, line 14: Where does this get used?

.github/workflows/sync-3rdparty-images.yaml
```diff
@@ -0,0 +1,104 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Log in to GitHub Docker Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Tag and push to ghcr
+        run: |
```
smoser, line 31: does 'run' run with 'set -e' by default?

.github/workflows/sync-3rdparty-images.yaml
```diff
@@ -0,0 +1,104 @@
...
+          docker trust inspect ${{ job.name }}:${{ matrix.version }}
+          docker pull ${{ job.name }}:${{ matrix.version }}
+          docker tag ${{ job.name }}:${{ matrix.version }} ghcr.io/${{ github.repository_owner }}/${{ job.name }}:${{ matrix.version }}
+          docker push ghcr.io/${{ github.repository_owner }}/${{ job.name }}:${{ matrix.version }}
```
smoser, line 35: Can't we just use one of these run sections and multiple name:versions ?

Or just something like this:

```bash
for n in ubuntu:latest alpine:edge centos:latest; do
    dest="${{ github.repository_owner }}/$n"
    docker trust inspect "$n"
    docker pull $n
    docker tag $n "$dest"
    docker push $dest
done
```

smoser (changes requested):
This feels liek too much text to accomplish what it is doing.  Can we just iterate over a loop of images to sync ? (see inline comment).

I don't actually think this will help much to remove dependency on docker.io.  
1. We still have 'docker://' references in several tests
2. we already were syncing these images to a local oci repo in the beginning of the build.  That sync is *still* using docker:// even after #544, so c-i builds still depend on it.

.github/workflows/sync-3rdparty-images.yaml
rchincha: Fixed like so ^

.github/workflows/sync-3rdparty-images.yaml
rchincha: Added it

raharper (commented):
LGTM  -- Can you resolve your issues? Or were you wanting @smoser to reply?

smoser (approved):

Issue #530, opened 2023-10-25
Bug: stacker does not notice import file changed


smoser:
### stacker version

v1.0.0-rc5-396ff9d liblxc cb8e38aca27a23964941f0f011a8919aab8bebab

### Describe the bug

When using import with a destination, stacker does not realize that a imported file has changed and re-build the content.

```yaml
demo:
  from:
    type: docker
    url: ${{IMAGE:docker://busybox:latest}}
  import:
    - path: import1.txt
      dest: /imports/
  run: |
    #!/bin/sh -e
    read line < /imports/import1.txt
    echo "read $line"
```

Given the above stacker.yaml, run `stacker build` once, change the content of 'import1.txt' and then run it again.  Expectation is that the build would be redone.

### To reproduce

I am attaching a script to automate the testing.
[stacker-test.sh.txt](https://github.com/project-stacker/stacker/files/13168917/stacker-test.sh.txt)

### Expected behavior

Build should occur a second time, after import1.txt content is changed and log shoudl contain output showing that it built.

rchincha:
HTTP HEAD request returns

< Content-Length: 34329287
< Last-Modified: Wed, 25 Oct 2023 02:50:40 GMT

and if available

rchincha:
<

smoser:
> HTTP HEAD request returns
> 
> < Content-Length: 34329287 < Last-Modified: Wed, 25 Oct 2023 02:50:40 GMT
> 
> and if available < Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=

To be clear, this is a local file.

Issue #516, opened 2023-09-29
Bug: No caching of import if dest is provided


smoser:
### stacker version

v1.0.0-rc5-396ff9d

### Describe the bug

Stacker does not cache any import items if a 'dest' is provided.

See example 'testme.sh'(https://github.com/project-stacker/stacker/files/12766496/testme.sh.txt) script here.

When executed, it will build a simple stacker.yaml file with 
```
  import:
    - path: https://github.com/cirros-dev/cirros/releases/download/0.6.2/cirros-0.6.2-source.tar.gz
      hash: 1bdd062118722501ed37234729d9164533da6bc3685f162c33924e64e36883b5
      dest: /imports/my.tar.gz
```

After doing so, it will look in the log for 'downloading'.  The idea is that the downloading of the url should only happen once.
That is true if you do not provide a 'dest' argument.  But if dest is provided then the url will be downloaded every time.

Issue #508, opened 2023-09-21
Feat: Allow user to specify a 'dest' to /stacker/imports


smoser:
### Is your feature request related to a problem? Please describe.

I'd like to be able to specify a 'dest' for a import that goes into /stacker/imports.
I think this is a significant shortcoming that should be addressed.  The workaround is for the user to specify a different path (not under /stacker, say '/imports'), but that results in 
a. having to consistently use that 'import' directory
b. having to clean up the import directory so it doesn't appear in the target.

Here is an example stacker.yaml 
```
build:
  build_only: true
  from:
    type: docker
    url: ${{DOCKER_BASE-docker://}busybox:latest
  import:
    - path: bar
      dest: bar.txt
    - path: foo
      dest: /stacker/imports/foo.txt
  run: |
    ls -l /stacker/imports/foo.txt /stacker/improts/bar.txt
```

Neither of the attempts above work with stacker version v1.0.0-rc5-396ff9d.

The first, using a relative path (bar), will complain:

    `error: Dest path cannot be relative for: map[interface {}]interface {}{"dest":"bar.txt", "path":"bar"}`

The second, using an absolute path but explicitly under /stacker/imports (foo), ends up populating an overlay that goes below /stacker/imports, so the 'run' section does not actually see it.

Here is an example script that i'd like to pass.
[test-case.sh.txt](https://github.com/project-stacker/stacker/files/12686753/test-case.sh.txt)

Issue #448, opened 2023-03-30
Feat: allow creation of layers via stacking (maybe bring back `apply`?)


smoser:
### Is your feature request related to a problem? Please describe.

I would like to be able to publish a set of individual layers and a "combined" layer, where the combined layer is joined layer of the individual layers.

An example, I have a stacker file like this:
```yaml
base:
  build_only: true
  from:
    type: docker
    url: docker://docker.io/library/ubuntu:jammy
  run:
    apt-get update -q

kernel-build:
  build_only: true
  from:
    type: built
    tag: base
  run: |
      mkdir -p /tmp/stage/kernel /export
      trap "rm -Rf /tmp/stage" EXIT

      apt-get install --no-install-recommends --assume-yes linux-image-virtual
      mv /boot /lib/modules /tmp/stage/kernel
      tar -C /tmp/stage -cf /export/kernel.tar kernel/

# the only top level dir this layer is 'kernel/'
kernel:
  from:
    type: scratch
    url: stacker://kernel-build/export/kernel.tar

ovmf-build:
  build_only: true
  from:
    type: built
    tag: base
  run: |
      mkdir -p /tmp/stage/ovmf /export
      trap "rm -Rf /tmp/stage" EXIT

      apt-get install --no-install-recommends --assume-yes ovmf
      cp /usr/share/OVMF/OVMF_CODE.secboot.fd /tmp/stage/ovmf/ovmf-code.fd
      cp /usr/share/OVMF/OVMF_VARS.fd /tmp/stage/ovmf/ovmf-vars.fd
      tar -C /tmp/stage -cf /export/ovmf.tar ovmf/

# the only top level dir this layer is 'ovmf/'
ovmf:
  from:
    type: scratch
    url: stacker://ovmf-build/export/ovmf.tar
```

The 'kernel' and 'ovmf' layers are built and have only top level directories 'kernel/' and 'ovmf/' respectively.

I'd like to add another layer build to this file that has:
```yaml
combined:
  from:
    type: stack
  layers:
    - type: built
      tag: kernel
    - type: built
      tag: ovmf
```

The goal would be to have 3 layers published:
 * ovmf
 * kernel
 * combined

combined would simply be 'ovmf' on top  of 'kernel', such that index.json had:

```json
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.oci.image.index.v1+json",
    "manifests": [
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
[...]

PR #548, opened 2023-11-13
feat(import): import container images directly

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


smoser (commented):
This is a similar feature to what i was asking for at #448 .

I don't think I like the use of 'import' for it though.  I'm not sure exactly why, but it seems a bit unnatural. 
Some questions this brings up to me:
 * Does every 'import' that specifies a 'dest' get its own layer? ie:
   ```yaml
   import:
     - path: file1.txt
       dest: /etc/file1.txt
     - path: file2.txt
       dest: /etc/file2.txt
   ```
     - Do those each get their own layer? I can see benefits to each approach. 
 * Why would this feature be limited to "single layer" container images?  Why shouldn't it "import and squish" or "import and *not* squish"?
 * I think this is just an optimization (admittedly useful) of function that is already there.  At least for the clean homebrew layout (where each layer contains its own paths without overlap) you could just create a 'ca-certificates' layer that was 'build_only: true' with 'from: docker / url: docker://ghcr.io/homebrew/core/ca-certificates:2022-10-11', and then 'import' the 'stacker://ca-certificates/ca-certificates' layer.

In related comment, i think import needs a fair amount of work (#456, #530, #516, #508).

rchincha:
> This is a similar feature to what i was asking for at #448 .
> 
> I don't think I like the use of 'import' for it though. I'm not sure exactly why, but it seems a bit unnatural. 

I would argue that we already import file paths and stacker:// and the only thing missing is docker://

Some questions this brings up to me:
> 
>     * Does every 'import' that specifies a 'dest' get its own layer? ie:

Yup.

>     * Why would this feature be limited to "single layer" container images?  Why shouldn't it "import and squish" or "import and _not_ squish"?

Intention is to initially keep this simple - 1:1 mapping between a layer and pkgname.

> 
>     * I think this is just an optimization (admittedly useful) of function that is already there.  At least for the clean homebrew layout (where each layer contains its own paths without overlap) you could just create a 'ca-certificates' layer that was 'build_only: true' with 'from: docker / url: docker://ghcr.io/homebrew/core/ca-certificates:2022-10-11', and then 'import' the 'stacker://ca-certificates/ca-certificates' layer.

True, but this is a general case where **all** such layers are not built in the same place and/or made available in the same registry.

> In related comment, i think import needs a fair amount of work (#456, #530, #516, #508).

Looks unrelated at least from code changes pov.

rchincha:
btw, there is definitely interest in the community wrt exactly this feature.

rchincha:
> Why would this feature be limited to "single layer" container images? Why shouldn't it "import and squish" or "import and not squish"?

This is for driving a specific use case of pkgs distributed as layers. Don't expect this to be a main use case immediately.
Also, we don't want to be changing any layers we import. That is a big no-no.

test/import.bats
```diff
@@ -541,6 +541,35 @@ newimports:
...
+
+compose-img:
+    from:
+      type: oci
+      url: $CENTOS_OCI
+    import:
+    - path: docker://ghcr.io/homebrew/core/openssl/1.1:1.1.1k
+      dest: /
+    - path: docker://ghcr.io/homebrew/core/curl:8.0.1
+      dest: /opt/
```
smoser, line 567: how does this work?
You take a layer that is rooted at "/" and place it at "/opt" ? Won't that have to create a new layer (sha256 blob tarball) , thus reducing the benefit and destroying any signatures on the imported layer?

smoser:
> homebrew is a pkg mgmt that distributes content via OCI layers, but apart from that it does more.

How is one supposed to use that?  As I showed, the layers are not usable.  They are missing dependencies.  What is supposed to resolve those dependencies? How is one *supposed* to use the curl oci layer?  Is it just left as an exercise to the user to find the correct layers to fulfill the 32 different missing layers?

raharper:
PR is quite old.  Please re-open when it's updated and ready.

PR #550, opened 2023-11-14
test: add code coverage support

rchincha:
With golang 1.20.x, we can build a binary with coverage support.

> Cover
>
> Go 1.20 supports collecting code coverage profiles for programs
> (applications and integration tests), as opposed to just unit tests.
> 
> To collect coverage data for a program, build it with go build's -cover
> flag, then run the resulting binary with the environment variable
> GOCOVERDIR set to an output directory for coverage profiles. See the
> 'coverage for integration tests' landing page for more on how to get
> started. For details on the design and implementation, see the
> proposal.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ BUILD_TAGS = exclude_graphdriver_btrfs exclude_graphdriver_devicemapper containe
 
 STACKER_OPTS=--oci-dir=$(BUILD_D)/oci --roots-dir=$(BUILD_D)/roots --stacker-dir=$(BUILD_D)/stacker --storage-type=overlay
 
-build_stacker = go build -tags "$(BUILD_TAGS) $1" -ldflags "-X main.version=$(VERSION_FULL) -X main.lxc_version=$(LXC_VERSION) $2" -o $3 ./cmd/stacker
+build_stacker = go build $1 -tags "$(BUILD_TAGS) $2" -ldflags "-X main.version=$(VERSION_FULL) -X main.lxc_version=$(LXC_VERSION) $3" -o $4 ./cmd/stacker
 
 # See doc/hacking.md for how to use a local oci or docker repository.
 STACKER_DOCKER_BASE?=docker://ghcr.io/project-stacker/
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -141,6 +158,21 @@ test: stacker $(REGCLIENT) $(SKOPEO) $(ZOT)
 		$(shell [ -z $(PRIVILEGE_LEVEL) ] || echo --privilege-level=$(PRIVILEGE_LEVEL)) \
 		$(patsubst %,test/%.bats,$(TEST))
 
+.PHONY: check-cov
+check-cov: lint test-cov
+
+.PHONY: test-cov
+test-cov: stacker-cov $(REGCLIENT) $(SKOPEO) $(ZOT)
```
smoser, line 165: is there ever a reason to call the unit tests without coverage?

smoser (commented):
this looks really good, thank you.  It will be really great to get "coverage credit" for the integration tests.

two overall comments:
1. Would it be simpler to just build both static-stacker and stacker-cov in the same stacker.yaml invocation? it messes up make dependencies a bit, but you can kind of just say "oh well" as stacker-cov always gets built when static-stacker is built.
2. Lets put the gocov into /stacker/working or something.  https://github.com/project-stacker/stacker/pull/543/files does a reasonable job of putting all our work under a single point inside.

unless there is some reason to *not* run integration adn unit tests with coverage, I really think we should just do that all the time.

Makefile
rchincha: We will make that call once we are able to get coverage data.

rchincha: We definitely want "go-test" (it does its own -cover etc) and bats tests separate.

rchincha: I would not optimize prematurely here. Let's see how this is working for us first.

rchincha:
Also, doesn't make sense to trust integration testing on an instrumented non-production binary - these should be kept separate.

pkg/container/userns.go
```diff
@@ -33,6 +33,12 @@ func MaybeRunInNamespace(config types.StackerConfig, userCmd []string) error {
 	euid := os.Geteuid()
 	env = append(env, fmt.Sprintf("%s=%d", envName, euid))
 
+	/*
```
smoser, line 36: we should drop this.

smoser (approved):
I have lots of comments on this, but none of them compare to a "+40% coverage" change.

I'll just put them here though for my later reference.

1. I don't think a separate "coverage" workflow is useful.  c-i should always run coverage. Im' not sure what caching we get between the 'coverage' and 'build' workflow, which  means just doing lots of extra things.  While they do seem to happen in parallel, that must multiplies the transient failure rate by 2. 
2. i think there are cleanups to be had in the makefile or stacker file changes.  we can do a better job there.

again, none of this comes near the *huge* win of getting coverage on integration tests.  thank you.

pkg/container/userns.go
rchincha: Fixed.

rchincha:
>I don't think a separate "coverage" workflow is useful. c-i should always run coverage. 

^ it is separate because binaries being built and shipped are separate (unfortunately) - we want the binary from one workflow and coverage from the other.

smoser:
> > I don't think a separate "coverage" workflow is useful. c-i should always run coverage.
> 
> ^ it is separate because binaries being built and shipped are separate (unfortunately) - we want the binary from one workflow and coverage from the other.

Just build them both from the same run.  A 'make' should build both. One is named 'stacker-coverage' one is named 'stacker'.

I can do a PR with my suggested cleanups at some point.

PR #551, opened 2023-11-15
test: Fix the docker-clone make target to exit failure.

smoser:
test: Fix the docker-clone make target to exit failure.

My expectation was that when make invoked the shell (sh) for
a line in a rule was that it did so with '-e'.  That was an
incorrect assumption.

The make docker-clone target was basically doing:

    for i in url1 url2 url3; do
        something $i; done

The failure of the *last* item in the list should cause the
command to exit failure, but failure of an earlier command would not.

The change here is just to move that shell code into a script in
tools/oci-copy and call it from there.


rchincha (approved):
lgtm

PR #555, opened 2023-11-18
fix: update go.mod to keep deps stable for compression behaviors

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


smoser (changes requested):
> fix: update go.mod to keep deps stable for compression behaviors 

Can you explain in your commit message more of what you're doing?

That is probably obvious to you right now, but its not obvious to me and I suspect it won't be obvious to a future-you either.

rchincha:
> > fix: update go.mod to keep deps stable for compression behaviors
> 
> Can you explain in your commit message more of what you're doing?
> 
> That is probably obvious to you right now, but its not obvious to me and I suspect it won't be obvious to a future-you either.

Will update.
We are using stacker, docker and skopeo together. And we want to pin some versions and keep the deps identical. This is not ideal until compression gets properly addressed upstream.

PR #557, opened 2023-11-20
test: Fix test failure due to .coverage changing during tar creation.

smoser:
This seems to be a transient failure, but is legitimate issue.

When coverage binary is being used, this test case could fail with:

    #  + mkdir /output
    # + tar -cf /output/installer-iso.tar -C /stacker .
    # tar: ./.coverage: file changed as we read it

The /stacker/.coverage file was being updated by the stacker binary itself.  This is just a race condition based on when the file gets written to and when it was being collected by tar.

The change is just to only have tar collect the .cpio files.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/test/dependency-order.bats
+++ b/test/dependency-order.bats
@@ -81,7 +81,7 @@ installer-iso-build:
         #!/bin/bash -ex
         # populate the iso
         mkdir /output
-        tar -cf /output/installer-iso.tar -C /stacker .
+        ( cd /stacker/imports && tar -cf /output/installer-iso.tar *.cpio )
 
 atomix-installer-iso:
     from:


rchincha (approved):

Makefile
```diff
@@ -171,7 +171,7 @@ test-cov: stacker-cov $(REGCLIENT) $(SKOPEO) $(ZOT)
 		STACKER_BUILD_UBUNTU_IMAGE=$(STACKER_BUILD_UBUNTU_IMAGE) \
 		./test/main.py \
 		$(shell [ -z $(PRIVILEGE_LEVEL) ] || echo --privilege-level=$(PRIVILEGE_LEVEL)) \
-		$(patsubst %,test/%.bats,$(TEST))
```
rchincha, line 174: As @hallyn pointed out, this needed?

smoser: fixed. good catch.

rchincha (approved):
lgtm

PR #558, opened 2023-11-20
ci: Only upload coverage when running coverage workflow, not build.

smoser:
I had a pull request (#557) that changed a single line in a single integration test.  c-i reported that coverage changed:
   5.72% (-48.53%) compared to a34ebfa
Where a34ebfa was the previous commit.

That doesn't sound right.  I think what is happening is that we have multiple uploads of the coverage workflow and they're fighting.

There isn't any reason to upload both anyway.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -97,8 +97,6 @@ jobs:
       - name: Show disk usage after running the tests
         if: always()
         uses: ./.github/actions/show-disk-usage
-      - name: Upload code coverage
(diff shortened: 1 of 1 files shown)


smoser:
the code coverage run showing the big change is https://app.codecov.io/gh/project-stacker/stacker/pull/557 .

rchincha:
Close this because https://github.com/project-stacker/stacker/pull/559

PR #559, opened 2023-11-20
ci: fix upload order

rchincha:
Uploading artifacts before code coverage appears to impact coverage stats. Perhaps there is a cleanup involved. So do it after.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


smoser:
@rchincha , Did you have any evidence behind your assumption that the order of upload caused the problem?

PR #564, opened 2023-11-27
Fix bom test failures

smoser:
* c-i: Un-mark alpine bom test as slow.
    
    Run the alpine bom test during normal c-i test.  This is a quick test
    (~10 seconds).  Having it run as part of c-i will mean that we sniff
    test the bom path without adding significant time.
    
  * c-i: Add alpine:edge image to those cloned and use it in test.
    
    The alpine image being used for the sbom test was already being
    used (and cloned).  There is no reason to depend on ghcr.io for it.
    
  * test: Fix bom tests to account for /stacker re-org
    
    Commit 2f284d8f401f0 re-organized the /stacker content layout.
    Instead of /stacker/tools/static-stacker we have /stacker/bin/stacker.
    

rchincha (approved):
lgtm

PR #567, opened 2023-11-30
feat: support importing sboms along with images

rchincha:
When we import an image via the from: directive, also pull in the sbom if there is one. Else we need to rebuild the sbom and there may not be enough state/information to do so.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/stacker/base.go
+++ b/pkg/stacker/base.go
@@ -46,7 +46,15 @@ func GetBase(o BaseLayerOpts) error {
 	case types.OCILayer:
 		fallthrough
 	case types.DockerLayer:
-		return importContainersImage(o.Layer.From, o.Config, o.Progress)
+		err := importContainersImage(o.Layer.From, o.Config, o.Progress)
+		if o.Layer.Bom != nil && o.Layer.Bom.Generate && (o.Layer.From.Type == types.DockerLayer) {
+			bomPath := path.Join(o.Config.StackerDir, "artifacts", o.Name)
+			err = getArtifact(bomPath, "application/spdx+json", o.Layer.From.Url, "", "", o.Layer.From.Insecure)
+			if err != nil {
+				log.Errorf("sbom for image %s not found", o.Layer.From.Url)
+			}
+		}
+		return err
 	default:
 		return errors.Errorf("unknown layer type: %v", o.Layer.From.Type)
 	}
--- a/pkg/stacker/publisher.go
+++ b/pkg/stacker/publisher.go
@@ -1,23 +1,14 @@
 package stacker
 
 import (
-	"bytes"
-	"context"
-	"encoding/base64"
-	"encoding/json"
 	"fmt"
 	"io"
-	"net/http"
-	"net/url"
 	"os"
 	"path"
 	"path/filepath"
 	"regexp"
 	"strings"
 
-	godigest "github.com/opencontainers/go-digest"
-	"github.com/opencontainers/image-spec/specs-go"
-	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/opencontainers/umoci"
 	"github.com/opencontainers/umoci/oci/casext"
 	"github.com/pkg/errors"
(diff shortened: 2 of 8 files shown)


pkg/stacker/referrer.go
```diff
@@ -0,0 +1,439 @@
...
+	slen := res.ContentLength
+	smtype := res.Header.Get("Content-Type")
+	sdgst, err := godigest.Parse(res.Header.Get("Docker-Content-Digest"))
+	if slen < 0 || smtype == "" || sdgst == "" || err != nil {
+		log.Errorf("unable to get descriptor details for subject:%s", subject)
+		if res.StatusCode != http.StatusNotFound {
+			return nil
+		}
+
+		return errors.Errorf("unable to get descriptor details for subject:%s", subject)
```
smoser, line 365: Maybe return a http.StatusNotFound or a specific error?

```go
ArtifactNotFound := errors.New("Artifact not found")
```

pkg/stacker/publisher.go
```diff
@@ -75,255 +66,6 @@ func NewPublisher(opts *PublishArgs) *Publisher {
...
-	res, err := http.DefaultClient.Do(req)
-	if err != nil {
-		log.Errorf("http request failed url:%s", url)
-		return nil, err
-	}
-
-	return res, nil
-}
-
-func fileDigest(path string) (*godigest.Digest, error) {
```
smoser, line 110: its strange to move this function 'fileDigest' from 'publisher.go' to 'referrer.go'.
Maybe you felt that it didn't really belong in "publisher.go", but to move it to another file with another name that doesn't really fit is strange.

Either leave it here (possibly to be cleaned up later) or move them to a better named file.

this could also be argued for clientRequest and publishArtifact  and 'basicAuth'

pkg/stacker/base.go
```diff
@@ -46,7 +46,15 @@ func GetBase(o BaseLayerOpts) error {
 	case types.OCILayer:
 		fallthrough
 	case types.DockerLayer:
-		return importContainersImage(o.Layer.From, o.Config, o.Progress)
+		err := importContainersImage(o.Layer.From, o.Config, o.Progress)
+		if o.Layer.Bom != nil && o.Layer.Bom.Generate {
+			bomPath := path.Join(o.Config.StackerDir, "artifacts", o.Name)
+			err = getArtifact(bomPath, "application/spdx+json", o.Layer.From.Url, "", "", o.Layer.From.Insecure)
+			if err != nil {
+				log.Warnf("sbom for image %s not found", o.Layer.From.Url)
```
smoser, line 54: "No SBOM" isn't really a warning.  If someone specified that a SBOM should be imported, then its an error.  If they didn't then its "operating as expected".

Can you turn that down to Info or even Debug?

test/bom.bats
```diff
@@ -1,5 +1,17 @@
 load helpers
 
+function setup_file() {
```
smoser, line 3: setup_file isn't supported in the version of bats that runs in c-i.  afaics c-i just installs 1.2.1 from the ubuntu 22.04 archive.

Even if it was, "zot_setup" is probably something you want to do from "suite_setup" (per test-suite run) rather than setup_file (per file).

I tried to do somethnig similar to this in while avoiding a newer bats dependency at https://github.com/project-stacker/stacker/pull/568/files . i dont' think it actually worked well though.

pkg/stacker/referrer_test.go
```diff
@@ -0,0 +1,27 @@
+package stacker
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestDistspecURLParsing(t *testing.T) {
```
smoser, line 8: thank you for unit tests.

smoser (commented):
there are some comments inline.

pkg/stacker/publisher.go
rchincha: Publishing involves two separate steps now.
Publish the image/tag with libcontainers APIs (old docker libs) and the new referrers where we have to handcraft HTTP requests. This has been moved into referrers.go now. fileDigest() was a helper routine for the second path.

pkg/stacker/referrer.go
```diff
@@ -0,0 +1,439 @@
...
+	subject := distspecManifestURL(registry, repo, subjectTag, skipTLS)
+
+	// check subject exists
+	res, err := clientRequest(http.MethodHead, subject, username, password, nil, nil)
+	if err != nil {
+		log.Errorf("unable to check subject:%s, err:%s", subject, err)
+		return err
+	}
+	if res == nil || res.StatusCode != http.StatusOK {
+		log.Errorf("subject:%s doesn't exist, ignoring and proceeding", subject)
```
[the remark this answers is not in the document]
rchincha: This is a situation where the original image may be gone, but the artifact (sbom, etc) may still be around, which is what we are interested in. We can catch the subject error in the `from:` section.

pkg/stacker/base.go
rchincha: Let's make this an error.

pkg/stacker/referrer.go
```diff
@@ -0,0 +1,439 @@
...
+
+	registry := durl.Host
+	repo := durl.Path
+	subjectTag := durl.Tag
+	subject := distspecManifestURL(registry, repo, subjectTag, skipTLS)
+
+	// check subject exists
+	res, err := clientRequest(http.MethodHead, subject, username, password, nil, nil)
+	if err != nil {
+		log.Errorf("unable to check subject:%s, err:%s", subject, err)
```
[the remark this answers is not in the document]
rchincha: either we log or wrap the specific error condition

hallyn (approved):
Thanks.

This looks good to me.

pkg/stacker/referrer.go
```diff
@@ -0,0 +1,442 @@
...
+
+	dgst, err := fileDigest(path)
+	if err != nil {
+		log.Errorf("unable get digest for file:%s, err:%s", path, err)
+		return err
+	}
+
+	fh, err := os.Open(path)
+	if err != nil {
+		log.Errorf("unable to open file:%s, err:%s", path, err)
```
raharper, line 161:
```suggestion
		log.Errorf("unable to open file:%q, err:%s", path, err)
```

pkg/stacker/referrer.go
```diff
@@ -0,0 +1,442 @@
...
+		return err
+	}
+	defer fh.Close()
+
+	if err := uploadBlob(registry, repo, path, username, password, fh, finfo.Size(), dgst, skipTLS); err != nil {
+		log.Errorf("unable to upload file:%s, err:%s", path, err)
```
raharper, line 167:
```suggestion
		log.Errorf("unable to upload file:%q, err:%s", path, err)
```

pkg/stacker/referrer.go
```diff
@@ -0,0 +1,442 @@
...
+		return err
+	}
+
+	// check and upload emptyJSON blob
```
raharper, line 171: check?  is this a check if we can upload?

pkg/stacker/referrer.go
```diff
@@ -0,0 +1,442 @@
...
+	// artifact manifest
+	mdgst := godigest.FromBytes(content)
+	regUrl := distspecManifestURL(registry, strings.Split(repo, ":")[0], mdgst.String(), skipTLS)
+	hdrs := map[string]string{
+		"Content-Type":   ispec.MediaTypeImageManifest,
+		"Content-Length": fmt.Sprintf("%d", len(content)),
+	}
+	res, err = clientRequest(http.MethodPut, regUrl, username, password, hdrs, bytes.NewBuffer(content))
+	if err != nil {
+		log.Errorf("unable to check subject:%s, err:%s", subject, err)
```
raharper, line 218:
```suggestion
		log.Errorf("unable to PUT subject:%s, err:%s", subject, err)
```
Here we're pushing the subject object, rright?

raharper: Or maybe we want to say, `failed to put artifact manifest` as the error here?

pkg/stacker/referrer.go
```diff
@@ -0,0 +1,442 @@
...
+
+func distspecBlobUploadURL(registry, repo string, skipTLS bool) string {
+	return fmt.Sprintf("%s://%s/v2%s/blobs/uploads/", distspecScheme(skipTLS), registry, strings.Split(repo, ":")[0])
+}
+
+func distspecReferrerURL(registry, repo, tag, artifactType string, skipTLS bool) string {
+	rurl := fmt.Sprintf("%s://%s/v2%s/referrers/%s", distspecScheme(skipTLS), registry, repo, tag)
+
+	aurl, err := url.Parse(rurl)
+	if err != nil {
```
raharper, line 255: can we at least log an error?

pkg/stacker/referrer.go
```diff
@@ -0,0 +1,442 @@
...
+	res, err := clientRequest(http.MethodHead, regUrl, username, password, nil, nil)
+	if err != nil {
+		log.Errorf("unable to check blob:%s, err:%s", subject, err)
+		return err
+	}
+	log.Debugf("HTTP response status:%v headers:%v", res.Status, dumpHTTPHeaders(res.Header))
+
+	hdr := res.Header.Get("Docker-Content-Digest")
+	if hdr != "" {
+		log.Infof("Copying blob %s skipped: already exists", dgst.Hex()[:12])
```
raharper, line 283: is there a reason to truncate the hash?

pkg/stacker/referrer.go
```diff
@@ -0,0 +1,442 @@
...
+		log.Errorf("unable to check subject:%s", subject)
+		return errors.Errorf("unable to check subject:%s", subject)
+	}
+
+	slen := res.ContentLength
+	smtype := res.Header.Get("Content-Type")
+	sdgst, err := godigest.Parse(res.Header.Get("Docker-Content-Digest"))
+	if slen < 0 || smtype == "" || sdgst == "" || err != nil {
+		log.Errorf("unable to get descriptor details for subject:%s", subject)
+		return errors.Errorf("unable to get descriptor details for subject:%s", subject)
```
raharper, line 362: This code exists else where, in `publishArtifact` can we make it common?

pkg/stacker/referrer.go
```diff
@@ -0,0 +1,442 @@
...
+	defer res.Body.Close()
+
+	var index ispec.Index
+	err = json.NewDecoder(res.Body).Decode(&index)
+	if err != nil {
+		return err
+	}
+
+	// we expect only one per artifactType
+	ref := index.Manifests[0].Digest
```
raharper, line 381: then should we complain if `len(index.Manifests) != 1` ?

raharper: we should probably check that so we avoid panic, no?

pkg/stacker/referrer.go
```diff
@@ -0,0 +1,442 @@
...
+	if (manifest.Config.MediaType != ispec.DescriptorEmptyJSON.MediaType) ||
+		(manifest.Config.Digest != ispec.DescriptorEmptyJSON.Digest) {
+		log.Errorf("invalid artifact descriptor for %s", sdgst.String())
+		return errors.Errorf("invalid artifact descriptor for %s", sdgst.String())
+	}
+
+	// create a tempfile
+	fh, err := os.CreateTemp(path, "*.json")
+	if err != nil {
+		log.Errorf("unable to open file:%s, err:%s", path, err)
```
raharper, line 406:
```suggestion
		log.Errorf("unable to open file:%q, err:%s", path, err)
```

pkg/stacker/referrer.go
```diff
@@ -0,0 +1,442 @@
...
+		return err
+	}
+	defer fh.Close()
+
+	// skipping additional OCI "artifact" checks
+	if err := downloadBlob(registry, repo, path, username, password, fh, manifest.Layers[0].Size, &manifest.Layers[0].Digest, skipTLS); err != nil {
+		log.Errorf("unable to download file:%s, err:%s", path, err)
```
raharper, line 413:
```suggestion
		log.Errorf("unable to download file:%q, err:%s", path, err)
```

pkg/stacker/referrer.go
```diff
@@ -0,0 +1,442 @@
...
+		return err
+	}
+
+	log.Infof("Copying artifact %s:%s done", path, mtype)
```
raharper, line 417:
```suggestion
	log.Infof("Copying artifact %q:%s done", path, mtype)
```

pkg/stacker/referrer.go
```diff
@@ -0,0 +1,442 @@
...
+
+	return nil
+}
+
+func downloadBlob(registry, repo, path, username, password string, writer io.Writer, size int64, dgst *godigest.Digest, skipTLS bool) error {
+	// upload with POST, PUT sequence
```
raharper, line 423: comment is stale

pkg/stacker/referrer_test.go
```diff
@@ -0,0 +1,27 @@
...
+	cases := map[string]*distspecUrl{
+		"docker://alpine:latest":                      &distspecUrl{Scheme: "docker", Host: "docker.io", Tag: "latest", Path: "/library/alpine"},
+		"docker://localhost:8080/alpine:latest":       &distspecUrl{Scheme: "docker", Host: "localhost:8080", Tag: "latest", Path: "/alpine"},
+		"docker://localhost:8080/a/b/c/alpine:latest": &distspecUrl{Scheme: "docker", Host: "localhost:8080", Tag: "latest", Path: "/a/b/c/alpine"},
```
raharper, line 12: what about:
```
"docker://localhost:8080/a/b/c/alpine:latest:withcolon": &distspecUrl{Scheme: "docker", Host: "localhost:8080", Tag: "latest:withcolon", Path: "/a/b/c/alpine"},

File: test/helpers.bash [hunk 21]
```diff
@@ -221,8 +222,11 @@ EOF
 }
 
 function zot_teardown {
+  echo "# stopping zot" >&3
   killall zot
+  killall -KILL zot || true
```
raharper, line 227: ooof -- why do we have to kill so hard?

PR #570, opened 2023-12-01
fix: Wrap use of skopeo to avoid its unwanted side effects.

smoser:
2 things are fixed here.
1. Running skopeo as root creates /run/containers with 700 permissions. That causes a problem when you run skopeo as root and then run stacker as non-root.

   The error would look like this:

   > error: initializing source ... getting username and password: 1 error occurred:
   >   * reading JSON file "/run/containers/0/auth.json":
   >         open /run/containers/0/auth.json: permission denied
   > couldn't import base layer base
   > stackerbuild.io/stacker/pkg/stacker.importContainersImage
   >       /stacker-tree/pkg/stacker/base.go:141
   > stackerbuild.io/stacker/pkg/stacker.GetBase
   >       /stacker-tree/pkg/stacker/base.go:49
   > stackerbuild.io/stacker/pkg/stacker.(*Builder).build
   >       /stacker-tree/pkg/stacker/build.go:407
   > stackerbuild.io/stacker/pkg/stacker.(*Builder).BuildMultiple
   >       /stacker-tree/pkg/stacker/build.go:622
   > main.doBuild

2. skopeo copy containers-registry populates ~/.local/share/containers/ This is an unwanted side effect of running the test.  Further annoying is that it ends up getting directories with 555 perms on them.  That means rm -Rf .local/share/containers/storage/vfs-layers will fail like:

      rm: cannot remove '.local/share/containers/storage/vfs/dir/HASH':
          Permission denied

   We don't want someone's HOME getting populated with artifacts from stacker test, and there are probably also race conditions here in that we run stacker tests in parallel.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -207,6 +207,24 @@ function zot_teardown {
   rm -f $TEST_TMPDIR/zot-config.json
 }
 
+function _skopeo() {
+    [ "$1" = "--version" ] && {
+        "$SKOPEO" "$@"
+        return
+    }
+    local uid=""
+    uid=$(id -u)
+    if [ ! -e /run/containers ]; then
+        if [ "$uid" = "0" ]; then
+            mkdir --mode=755 /run/containers || chmod /run/containers 755
+        fi
+    fi
+    [ -n "$TEST_TMPDIR" ]
+    local home="${TEST_TMPDIR}/home"
+    [ -d "$home" ] || mkdir -p "$home"
+    HOME="$home" "$SKOPEO" "$@"
+}
+
 function test_copy_buffer_size() {
    local buffer_size=$1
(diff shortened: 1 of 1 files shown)


test/helpers.bash
```diff
@@ -261,5 +279,5 @@ EOF
   stacker clean
   rm -rf folder1
   cd "$ROOT_DIR"
-  rm -rf "tmpdir"
+  rm -rf "$tmpdir"
```
raharper, line 282: worth mentioning you fixed a case where we left a tmpdir around when using `test_copy_buffer_size` I think.

rchincha (approved):
lgtm

Issue #571, opened 2023-12-04
`import` directive is deprecated and replaced by `imports`.


smoser:
### Deprecation of `import` directive

The `import` directive in `stacker.yaml` files is deprecated.  It is replaced by `imports` (plural).

**Support for `import` will not be present in any stacker release made after 2025-01-01**.

The `import` directive contained all the functionality of the `imports` directive in stacker releases up until 0.4.0 and initial release of the 1.0 branch.  1.0 pre-releases contain a warning on use of `import`.  Users must migrate to `imports` directive.

### Migration path
In order to change from `import` to `imports` the user will:

1. change use of `import` in stacker.yaml files to `imports`
2. change reference to imported items from within a `run` section from `/stacker/<path>` to `/stacker/imports/<path>`. 

As an example, consider this old `stacker-old.yaml` file:

```yaml
build:
  from:
    type: docker
    url: ${{DOCKER_BASE:docker://}}busybox:latest
  build_only: true
  import:
    - path: tools/fizz-buzz
    - path: tools/frobnicator
      dest: /opt/bin/frobnicator
      perms: 0755
  run: |
    cp /stacker/fizz-buzz /opt/bin/fizz-buzz
    chmod 755 /opt/bin/fizz-buzz
```

Building the above `stacker-old.yaml` file will give a deprecation warning.
To use `imports`, the new `stacker.yaml` would look like this:

```yaml
build:
  from:
    type: docker
    url: ${{DOCKER_BASE:docker://}}busybox:latest
  build_only: true
  imports:
    - path: tools/fizz-buzz
    - path: tools/frobnicator
      dest: /opt/bin/frobnicator
      perms: 0755
  run: |
    cp /stacker/imports/fizz-buzz /opt/bin/fizz-buzz
    chmod 755 /opt/bin/fizz-buzz
```

The changes are highlighted in this diff:

```diff
@@ -3,11 +3,11 @@
     type: docker
     url: ${{DOCKER_BASE:docker://}}busybox:latest
   build_only: true
-  import:
+  imports:
     - path: tools/fizz-buzz
     - path: tools/frobnicator
       dest: /opt/bin/frobnicator
       mode: 0755
   run: |
-    cp /stacker/fizz-buzz /opt/bin/fizz-buzz
+    cp /stacker/imports/fizz-buzz /opt/bin/fizz-buzz
     chmod 755 /opt/bin/fizz-buzz
```

### Reason for change
[...]

PR #572, opened 2023-12-04
fix: Update 'import' deprecation warning with a date.

smoser:
Putting a date in the deprecation warning message allows users to plan migration more effectively.  The date chosen allows for plenty of time for users to migrate existing stacker.yaml files.

#571.

### DIFF
--- a/pkg/types/stackerfile.go
+++ b/pkg/types/stackerfile.go
@@ -220,8 +220,11 @@ func NewStackerfile(stackerfile string, validateHash bool, substitutions []strin
 	for _, name := range sf.FileOrder {
 		layer := sf.internal[name]
 		if layer.WasLegacyImport() {
-			log.Warnf("Deprecated 'import' directive used in layer '%s' of file '%s'. "+
-				"Change from 'import' to 'imports', and '/stacker/' to '/stacker/imports/'", name, stackerfile)
+			log.Warnf("'import' directive used in layer '%s' inside file '%s' is deprecated. "+
+				"Support for 'import' will be removed in releases after 2025-01-01. "+
+				"Migrate by changing 'import' to 'imports' and '/stacker' to '/stacker/imports'. "+
+				"See https://github.com/project-stacker/stacker/issues/571 for migration.",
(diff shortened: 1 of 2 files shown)


pkg/types/stackerfile.go
```diff
@@ -220,8 +220,10 @@ func NewStackerfile(stackerfile string, validateHash bool, substitutions []strin
 	for _, name := range sf.FileOrder {
 		layer := sf.internal[name]
 		if layer.WasLegacyImport() {
-			log.Warnf("Deprecated 'import' directive used in layer '%s' of file '%s'. "+
-				"Change from 'import' to 'imports', and '/stacker/' to '/stacker/imports/'", name, stackerfile)
+			log.Warnf("'import' directive used in layer '%s' inside file '%s' is deprecated. "+
```
raharper, line 223: You dropped the `Change from 'import' -> 'imports'` which I think is critical to allowing folks to fix things ASAP without loading up a webpage.  I'd keep the original line 223/224, and then add your new lines.

raharper (changes requested):

pkg/types/stackerfile.go
smoser: The message occurs every time a layer is loaded. its kind of annoying.
I figure the url is a place to document how to fix it.  

> "Change from 'import' to 'imports', and '/stacker/' to '/stacker/imports/'"

probably wasn't completely clear to anyone that wasn't familiar anyway.

If people think we should put that back in, i'm fine to do so... its just that the message is so long.

rchincha: For something we are deprecating, best to complain?

smoser: done.

smoser:
i'm not actually sure why [ df405f7 ](https://github.com/project-stacker/stacker/commit/df405f71e9cd4876a1900a07e1fed5e2dfb299e7) failed and 2dd1369 succeeded. :-(

rchincha (approved):
lgtm

PR #577, opened 2023-12-12
fix: use overlay whiteouts when extracting tar layers

rchincha:
If a tar layer has a deleted file, a whiteout entry is created. However, when extracting tar and overlay whiteouts are incompatible.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/overlay/pack.go
+++ b/pkg/overlay/pack.go
@@ -725,7 +725,9 @@ func unpackOne(l ispec.Descriptor, ociDir string, extractDir string) error {
 			return err
 		}
 
-		err = layer.UnpackLayer(extractDir, uncompressed, nil)
+		// always unpack with Overlay whiteout mode to prevent ignoring whiteouts in tar layers
+		// see test/publish.bats: "building from published images with whiteouts" for more details
+		err = layer.UnpackLayer(extractDir, uncompressed, &layer.UnpackOptions{WhiteoutMode: layer.OverlayFSWhiteout})
 		if err != nil {
 			if rmErr := os.RemoveAll(extractDir); rmErr != nil {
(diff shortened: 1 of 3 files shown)


pkg/overlay/pack.go
```diff
@@ -725,7 +725,7 @@ func unpackOne(l ispec.Descriptor, ociDir string, extractDir string) error {
 			return err
 		}
 
-		err = layer.UnpackLayer(extractDir, uncompressed, nil)
+		err = layer.UnpackLayer(extractDir, uncompressed, &layer.UnpackOptions{WhiteoutMode: layer.OverlayFSWhiteout})
```
raharper, line 728:
```suggestion
                // always unpack with Overlay whiteout mode to prevent ignoring whiteouts in tar layers
                // see test/publish.bats: "building from published images with whiteouts" for more details
		err = layer.UnpackLayer(extractDir, uncompressed, &layer.UnpackOptions{WhiteoutMode: layer.OverlayFSWhiteout})
```

raharper: Feel free to expand on the comment; I like to seem context comments when making optional choices like this one.

test/publish.bats
```diff
@@ -243,3 +243,28 @@ published:
 EOF
 
 }
+
+@test "building from published images with whiteouts" {
```
raharper, line 247: Could we describe the test and what we expect and the observed failure here in comments?

raharper (commented):
Nice find and straight-forward fix.  I added some requests for comments, see what you think.

pkg/overlay/pack.go
rchincha: Added.

PR #578, opened 2023-12-14
Switch idmap from lxd to incus

hallyn:
Incus had moved idmap from shared/ to internal/.

This is a draft pr to test with my custom incus which moves it back.


rchincha (approved):
lgtm

PR #579, opened 2023-12-14
fix: catch failures in makefile when building skopeo

smoser:
Skopeo's build target is an inline shell script.
It needs to run with 'set -e' or it will just keep going after it fails.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,7 @@ $(ZOT):
 	$(call dlbin,$@,https://github.com/project-zot/zot/releases/download/$(ZOT_VERSION)/zot-linux-amd64-minimal)
 
 $(SKOPEO):
-	@mkdir -p "$(TOOLS_D)/bin"; \
+	@set -e; mkdir -p "$(TOOLS_D)/bin"; \
 	tmpdir=$$(mktemp -d); \
 	cd $$tmpdir; \
 	git clone https://github.com/containers/skopeo.git; \


Makefile
```diff
@@ -126,7 +126,7 @@ $(ZOT):
 	$(call dlbin,$@,https://github.com/project-zot/zot/releases/download/$(ZOT_VERSION)/zot-linux-amd64-minimal)
 
 $(SKOPEO):
-	@mkdir -p "$(TOOLS_D)/bin"; \
+	@set -e; mkdir -p "$(TOOLS_D)/bin"; \
```
raharper, line 129: This looks fine to me, though could use commands joined with `&&` instead of `set -e`.

mikemccracken:
would this be better using `.ONESHELL:` and `.SHELLFLAGS: -e` ?https://www.gnu.org/software/make/manual/html_node/One-Shell.html

Makefile
smoser: there really isn't any point in 'set -e' *and* '&&' . they're kind of redundant. no?

smoser:
I think .ONESHELL and .SHELLFLAGS are wierd (they're like decorators on a rule) but i'm fine to switch to that if other people want that.  

it seems like `set -e` woudl have been a batter choice as the default behavior.

Makefile
raharper: You're right; I suggested to use `&&` instead since we've discussed the short-comings of `set -e` elsewhere.   Here I think it works fine.

smoser:
I thought I had commented, if people think ONESHELL and SHELLFLAGS make this better, i'm fine to do that.

I guess either ACK the fix here or request changes and I'll comply.

raharper (approved):

PR #580, opened 2024-01-05
fix: don't break on --substitute values which have commas

hallyn:
The following:

  stacker build --substitute a=b,c

used to work, until commit 3baba644:

  fix(ci): convert ci is failing due to perms (#439)

This squashed PR included 'chore(go.mod): update cli to github.com/urfave/cli/v2'.  Switching from urfave/cli/v1 to v2 changes string slice flag behavior to automatically split on commas.

Luckily there is an app.DisableSliceFlagSeparator flag we can set to tell it not to do that.  Set that flag.  And add a test for this.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/test/basic.bats
+++ b/test/basic.bats
@@ -304,3 +304,15 @@ EOF
     [ ! -f dest/rootfs/favicon.ico ]
     [ ! -d dest/rootfs/stacker ]
 }
+
+@test "commas in substitute flags ok" {
+    cat > stacker.yaml <<EOF
+busybox:
+    from:
+        type: oci
+        url: $BUSYBOX_OCI
+    run: |
+        touch /foo
+EOF
+    stacker build --substitute "a=b,c"
+}
--- a/cmd/stacker/main.go
+++ b/cmd/stacker/main.go
@@ -123,6 +123,8 @@ func main() {
 		&checkCmd,
 	}
 
+	app.DisableSliceFlagSeparator = true
+
 	app.Flags = []cli.Flag{
 		&cli.StringFlag{
 			Name:  "work-dir",


hallyn:
> not ok 66 import various sizes in 4sec

@rchincha is this a regression in this test?  (i don't see how it could be, but...)  Or has it been seen elsewhere?

test/basic.bats
```diff
@@ -304,3 +304,15 @@ EOF
...
+@test "commas in substitute flags ok" {
+    cat > stacker.yaml <<EOF
+busybox:
+    from:
+        type: oci
+        url: $BUSYBOX_OCI
+    run: |
+        touch /foo
+EOF
+    stacker build --substitute "a=b,c"
```
rchincha, line 317: How is this substitution used though in the stacker.yaml?

hallyn: Do you mean in the testcase?  It's not.  That's not necessary to trigger the regression failure-to-build in main branch.

hallyn: Hm. When I run this test locally (sudo bats test/basic.bats), i get zero failures.

hallyn: No, I was wrong - I can reproduce this.  No idea why.

rchincha: The CI failure is due to the gzip issue. It is a concern, but unrelated to this PR.

hallyn: Actually it seems to just be a flaky test? :   

```
ubuntu@stacker:~/stacker$ sudo -E bats test/gzip.bats 
 ✓ import various sizes 

1 test, 0 failures
```

PR #581, opened 2024-01-10
ci: update bats and zot versions

rchincha:
Update Makefile to bump up zot version
Update Makefile to include a recipe to install latest bats

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,6 @@ LXC_BRANCH?=stable-5.0
 HACK_D := $(TOP_LEVEL)/hack
 # helper tools
 TOOLS_D := $(HACK_D)/tools
-export PATH := $(TOOLS_D)/bin:$(PATH)
 REGCLIENT := $(TOOLS_D)/bin/regctl
 REGCLIENT_VERSION := v0.5.1
 SKOPEO = $(TOOLS_D)/bin/skopeo
@@ -142,7 +141,7 @@ $(SKOPEO):
 	rm -rf $$tmpdir;
 
 $(BATS):
-	rm -rf bats-core; \
+	@set -e; rm -rf bats-core; \
 	git clone -b $(BATS_VERSION) https://github.com/bats-core/bats-core.git; \
 	cd bats-core; ./install.sh $(TOOLS_D); cd ..; \
 	rm -rf bats-core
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -41,9 +41,11 @@ REGCLIENT := $(TOOLS_D)/bin/regctl
 REGCLIENT_VERSION := v0.5.1
 export SKOPEO = $(TOOLS_D)/bin/skopeo
 export SKOPEO_VERSION = 1.9.3
+export BATS = $(TOOLS_D)/bin/bats
```
smoser, line 44: there is no reason to export this if you never use it.
just drop the 'export'

smoser (changes requested):
suggested changes in https://github.com/rchincha/stacker/pull/10

Makefile
```diff
@@ -37,13 +37,16 @@ LXC_BRANCH?=stable-5.0
 HACK_D := $(TOP_LEVEL)/hack
 # helper tools
 TOOLS_D := $(HACK_D)/tools
+export PATH := $(TOOLS_D)/bin:$(PATH)
```
rchincha, line 40: Does this even work with sudo -E?

Makefile
```diff
@@ -164,10 +171,8 @@ check-cov: lint test-cov
 
 .PHONY: test-cov
 test-cov: stacker-cov $(REGCLIENT) $(SKOPEO) $(ZOT)
-	sudo -E PATH="$$PATH" \
```
smoser, line 167: so we do need this.

Makefile
smoser:
```
$ export PATH=/tmp/foo:$PATH ; sudo -E sh -c 'echo $PATH'
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
```
[sudo(8)](https://manpages.ubuntu.com/manpages/jammy/en/man8/sudo.8.html) says:

>  PATH             May be overridden by the security policy.

So i guess this doesn't work at least in the default ubuntu path here.

So I guess we have to leave PATH=PATH.

Makefile
```diff
@@ -147,10 +155,8 @@ PRIVILEGE_LEVEL?=
 check: lint test go-test
 
 .PHONY: test
-test: stacker $(REGCLIENT) $(SKOPEO) $(ZOT)
-	sudo -E PATH="$$PATH" \
-		LXC_BRANCH=$(LXC_BRANCH) \
-		LXC_CLONE_URL=$(LXC_CLONE_URL) \
+test: stacker $(REGCLIENT) $(SKOPEO) $(ZOT) $(BATS)
```
smoser, line 158: because test is a .PHONY dependencies don't matter and aren't resolved.
nothing ends up building bats and so test fails.

https://stackoverflow.com/questions/13852535/makefile-dependencies-dont-work-for-phony-target

smoser:
https://github.com/project-stacker/stacker/pull/584 has a fix that makes this pass.

PR #586, opened 2024-01-22
fix: use a umoci which has fix for another empty lgetxattr

hallyn:
We end up with errors like:

/home/shallyn/stackertest2/roots/test2/overlay/etc/systemd/system/multi-user.target.wants: dropped overlay attrs: user.overlay.origin error: no data available
get xattr: user.overlay.origin
github.com/opencontainers/umoci/oci/layer.(*tarGenerator).AddFile
        /stacker-tree/.build/gopath/pkg/mod/github.com/project-stacker/umoci@v0.0.0-20231025233220-c7bb24d57e98/oci/layer/tar_generate.go:225
github.com/opencontainers/umoci/oci/layer.GenerateInsertLayer.func1.2

Unless there is an actual data i/o error, ignore it.  In particular, if the fs either says eopnotsupp, enodata, or eperm, just assume the xattr did not actually exist.  We have to do this because there are cases where listxattr will return xattr keys which getxattr will filter out, and we're expected to ignore those.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


hallyn:
I'm using my tree so builds can happen.  Obviously we'll want to get that fix back into github.com/project-stacker/umoci

go.sum
```diff
@@ -508,6 +508,8 @@ github.com/gopherjs/gopherjs v0.0.0-20181017120253-0766667cb4d1/go.mod h1:wJfORR
 github.com/gorilla/mux v1.8.1 h1:TuBL49tXwgrFYWhqrNgrUNEY92u81SPhu7sTdzQEiWY=
 github.com/gorilla/mux v1.8.1/go.mod h1:AKf9I4AEqPTmMytcMc0KkNouC66V3BtZ4qD5fmWSiMQ=
 github.com/grpc-ecosystem/grpc-gateway v1.16.0/go.mod h1:BDjrQk3hbvj6Nolgz8mAMFbcEtjT1g+wF4CSlocrBnw=
+github.com/hallyn/umoci v0.0.0-20240122192320-3c661c1beb44 h1:4xaZ4I8PCnycSDr0RHn0JJVJtuJr7Q852ui8ChOI8rQ=
```
andaaron, line 511: Run a go mod tidy. Should update these lines and fix the compilation errors.

hallyn: Will do, but note that I had done a go mod tidy, a full build, and a full test locally yesterday.  Seems fragile...

hallyn: Done, hoping for the best, thanks :)

rchincha (approved):
lgtm

PR #598, opened 2024-02-28
2024.02.27/main/stricter substitutions

mikemccracken:
**What type of PR is this?**

This is a breaking change to the substitution syntax, which was bad and should be made good.

This makes it an error to have placeholders that are valid shell like `$PATH` or `${PATH}` if you have provided a substitution for `PATH` to stacker via `--substitute PATH=why:oh:why`.

Those placeholders are too easy to confuse with a valid shell variable, and stacker can't tell if you intended to provide a value for them but typoed it - and you can't tell by looking at it if it is intended to be a stacker sub or just a shell var.

So they are bad and should be made good by becoming fatal errors.

The documented behavior in docs was not tested fully, and was in fact wrong. Now it's right.

This change is confined to one function that has go unit tests, so I just updated those and the one place where our tests were using `$PLAIN` syntax.

Possibly. If people were doing confusing things like `--substitute PATH=ouch`, then their builds will fail with a helpful message about how to fix it.

```release-note
Substitution placeholders in stacker files with no braces or single braces are no longer supported, and an error will be printed to explain how to replace them. 

all valid shell variables with no braces or single braces are still fine, the change is that now if you provide a substitution `--substitute FOO=val`, then use `$FOO` or `${FOO}` in your stacker file, that will be a fatal error. You must use `${{FOO}}` instead to have stacker replace it with the provided substitution.
```

An example of the new error output is:
```
2024/02/27 16:36:03 error "A=B" was provided as a substitution and unsupported placeholder "${A}" was found. Replace "${A}" with "${{A}}" to use the substitution.
 isStacker=&{}
```

### DIFF
--- a/doc/stacker_yaml.md
+++ b/doc/stacker_yaml.md
@@ -11,19 +11,35 @@ processed in stacker as:
 
     1 2 3
 
-That is, variables of the form `$FOO` or `${FOO}` are supported, and variables
-with `${FOO:default}` a default value will evaluate to their default if not
-specified on the command line. It is an error to specify a `${FOO}` style
-without a default; to make the default an empty string, use `${FOO:}`.
+In order to avoid conflict with bash or POSIX shells in the `run` section, 
+only placeholders with two braces are supported, e.g. `${{FOO}}`.
+Placeholders with a default value like `${{FOO:default}}` will evaluate to their default if not
+specified on the command line or in a substitution file.
 
-In addition to substitutions provided on the command line, the following
-variables are also available with their values from either command
+Using a `${{FOO}}` placeholder without a default will result in an error if
+there is no substitution provided. If you want an empty string in that case, use
+an empty default: `${{FOO:}}`.
+
+In order to avoid confusion, it is also an error if a placeholder in the shell
+style (`$FOO` or `${FOO}`) is found when the same key has been provided as a
+substitution either via the command line (e.g. `--substitute FOO=bar`) or in a
+substitution file. An error will be printed that explains how to rewrite it:
+
+   error "A=B" was provided as a substitution and unsupported placeholder "${A}" was found. Replace "${A}" with "${{A}}" to use the substitution.
+
+Substitutions can also be specified in a yaml file given with the argument
+`--substitute-file`, with any number of key: value pairs:
+
+    FOO: bar
+    BAZ: bat
+
+In addition to substitutions provided on the command line or a file, the
+following variables are also available with their values from either command
 line flags or stacker-config file.
 
     STACKER_STACKER_DIR config name 'stacker_dir', cli flag '--stacker-dir'-
     STACKER_ROOTFS_DIR  config name 'rootfs_dir', cli flag '--roots-dir'
     STACKER_OCI_DIR     config name 'oci_dir', cli flag '--oci-dir'
-
 
 The stacker build environment will have the following environment variables
 available for reference:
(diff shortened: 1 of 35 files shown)


mikemccracken:
Note that there are two commits. The first commit removes support for the other sub styles, but just warns when they are found to be shadowed. The second commit upgrades them to errors, which I think is the right way to go, but am open to discussion.

rchincha:
# "BUSYBOX_OCI=/home/runner/work/stacker/stacker/test/busybox:latest" was provided as a substitution and unsupported placeholder "$BUSYBOX_OCI" was found. Replace "$BUSYBOX_OCI" with "${{BUSYBOX_OCI}}" to use the substitution.
 
 ^ breaking test.

rchincha (approved):
lgtm

Have to make changes to our official docs.

doc/stacker_yaml.md
```diff
@@ -11,20 +11,28 @@ processed in stacker as:
...
-specified on the command line. It is an error to specify a `${FOO}` style
-without a default; to make the default an empty string, use `${FOO:}`.
+In order to avoid conflict with bash or POSIX shells in the `run` section, 
+only placeholders with two braces are supported, e.g. `${{FOO}}`.
+Placeholders with a default value like `${{FOO:default}}` will evaluate to their default if not
+specified on the command line or in a substitution file.
 
-In addition to substitutions provided on the command line, the following
-variables are also available with their values from either command
+It is an error to specify a `${{FOO}}` placeholder without a default; to make
```
hallyn, line 19: Wouldn't it be just as robust and more intuitive to have ${{FOO}} work fine when FOO is set, and to simply fail (with a good error message) if FOO substitution is not defined?  Then ${{FOO:}} would still be needed if you want to default to empty, but i don't believe that's often the intent.

mikemccracken: That's the behavior you get now. If you have `${{FOO}}` and you don't provide `--substitution FOO=whatever`, you will get the error `No value for substitution FOO`.

mikemccracken: (so this does need to be worded more clearly, I'll do that)

mikemccracken: (done)

pkg/types/stackerfile.go
```diff
@@ -58,28 +58,69 @@ func (sf *Stackerfile) Len() int {
...
 		to := membs[1]
 
-		log.Debugf("substituting %s to %s", from, to)
+		// warn on finding unsupported placeholders matching provided
+		// substitution keys:
+		nobracket := fmt.Sprintf("$%s", membs[0])
+		onebracket := fmt.Sprintf("${%s}", membs[0])
+		bad_content := content
+		for _, unsupp_placeholder := range []string{nobracket, onebracket} {
+			bad_content := strings.Replace(content, unsupp_placeholder, to, -1)
```
hallyn, line 77: Wouldn't strings.Index(unsupp_placeholder) != -1 work as well and be faster?

mikemccracken:
I noticed that I hadn't limited the test changes to the last commit, so I just reworked the last two.
The test changes should work even without the other commits, so we could even consider putting them first, if we want every commit to be able to pass CI on its own

pkg/types/stackerfile.go
mikemccracken: yep

mikemccracken: I changed this in the second commit just now

hallyn (approved):
Thanks.

hallyn:
Did you want to squash these (to combine the commit msgs yourself), or just squash-and-merge?

hallyn (approved):

PR #602, opened 2024-03-18
add stacker version as annotation to images

mikemccracken:
It can be useful to know which version of stacker created an image. Let's put it in there.

Moves some things around to avoid multiple definitions & import loops.

### DIFF
--- a/pkg/stacker/build.go
+++ b/pkg/stacker/build.go
@@ -18,6 +18,7 @@ import (
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
 	"stackerbuild.io/stacker/pkg/container"
+	"stackerbuild.io/stacker/pkg/lib"
 	"stackerbuild.io/stacker/pkg/log"
 	"stackerbuild.io/stacker/pkg/test"
 	"stackerbuild.io/stacker/pkg/types"
@@ -281,6 +282,7 @@ func (b *Builder) updateOCIConfigForOutput(sf *types.Stackerfile, s types.Storag
 	}
 
 	annotations[getStackerContentsAnnotation(opts.AnnotationsNamespace)] = sf.AfterSubstitutions
+	annotations[getStackerVersionAnnotation(opts.AnnotationsNamespace)] = lib.StackerVersion
 
 	history := ispec.History{
 		EmptyLayer: true, // this is only the history for imageConfig edit
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,8 @@ BUILD_TAGS = exclude_graphdriver_btrfs exclude_graphdriver_devicemapper containe
 
 STACKER_OPTS=--oci-dir=$(BUILD_D)/oci --roots-dir=$(BUILD_D)/roots --stacker-dir=$(BUILD_D)/stacker --storage-type=overlay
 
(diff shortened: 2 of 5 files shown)


pkg/stacker/build.go
```diff
@@ -281,6 +282,7 @@ func (b *Builder) updateOCIConfigForOutput(sf *types.Stackerfile, s types.Storag
 	}
 
 	annotations[getStackerContentsAnnotation(opts.AnnotationsNamespace)] = sf.AfterSubstitutions
+	annotations[getStackerVersionAnnotation(opts.AnnotationsNamespace)] = lib.StackerVersion
```
raharper, line 285: What does the output look like in the OCI?

hallyn:
Looks good to me, but would also like to see the oci annotation.

mikemccracken:
here's an example of a manifest with the annotation we get from this change:
```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "config": {
    "digest": "sha256:8234d3f66d0841bdaf46c406fa657ff0c8656a140cc9644169f6437268668ff1",
    "size": 709
  },
  "layers": [
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
      "digest": "sha256:6d7944e47a2816085f3558cbd8f8b1c9ab7ea946908701dc86e824d41316daf4",
      "size": 1916744
    },
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
      "digest": "sha256:8f8602a5f876184701dc7aff30b2401f8cd6b1cf80153ec560c9b6b1927041aa",
      "size": 209
    }
  ],
  "annotations": {
    "io.stackeroci.stacker.stacker_version": "v1.0.0-rc12-3-gef4e4be-dirty",
    "io.stackeroci.stacker.stacker_yaml": "\nminbasebase:\n  build_only: true\n  from:\n    type: tar\n    url: minrfs-w-resolvconf-and-dirs.tgz\n\nminbase:\n  from:\n    type: built\n    tag: minbasebase\n  run: |\n    echo \"foo\" > \"/usr/local/bin/foo\"\n    echo \"$(</etc/resolv.conf)\"\n    echo done.\n"
  }
}
```

hallyn (approved):

raharper (approved):
+1 -- maybe a quick pkg/lib/version_test.go to make codecov happy?

mikemccracken:
> +1 -- maybe a quick pkg/lib/version_test.go to make codecov happy?

I'm not too familiar with codecov, having a hard time making sense of its view. I don't think it's referring to pkg/lib/version.go - it claims the lines added in api.go don't have tests associated, but there are no unit tests for anything in there (and there shouldn't be, it is just wrappers around Sprintf)... I assume it's getting coverage from the integration bats tests too, but when I look in codecov to see how the other funcs in api.go are being covered, it now says all the lines in that file are covered - in this view of the PR on codecov https://app.codecov.io/gh/project-stacker/stacker/pull/602 it says the PR's changes have 100% coverage. Not sure how to satisfy it

raharper:
> > +1 -- maybe a quick pkg/lib/version_test.go to make codecov happy?
> 
> I'm not too familiar with codecov, having a hard time making sense of its view. 

I'm not quite sure either, but most files in `pkg/lib/ have `_test.go` files -- could write one and see if that' makes it happy?

rchincha (approved):
all this should be part of in-toto metadata in the long-term

PR #603, opened 2024-03-20
fix: cache: serialize legacy import field

### LINKED ISSUE #592: Bug: the 'import' fallback fails for 'prerequisites'
(This bug was found by @ariel-miculas and @andaaron was the one who guessed at the 'import' being the problem)

### stacker version

v1.0.0-rc11

### Describe the bug

'import:' in stacker.yaml has been replaced by 'imports:'.  Those files end up under /stacker/imports/ instead of directly under /stacker.  In order to provide some backward compatible support as users switch over, there is fallback support for 'import:' which is scheduled to be removed in the future.

This support however breaks mysteriously when using the 'prerequisites' feature.

### To reproduce

1. Configuration

```
cat >> stacker-req.yaml << EOF
base:
    from:
      type: docker
      url: docker://busybox:latest
    build_only: true
    import:
        - helloworld
    run: |
      cp /stacker/helloworld /

EOF
cat >> stacker.yaml << EOF
config:
  prerequisites:
      - stacker-req.yaml

buildenv:
    from:
        type: built
        tag: base
    import:
        - stacker://base/helloworld
    run: |
        cp /stacker/helloworld /h3

EOF
stacker build
```
2. Client tool used
The latest stacker pre-release.
4. Seen error
```
serge@serge-l-PF3DENS3 ~/delme/stacker-imports$ ./stacker build
'import' directive used in layer 'buildenv' inside file 'stacker.yaml' is deprecated. Support for 'import' will be removed in releases after 2025-01-01. Migrate by changing 'import' to 'imports' and '/stacker' to '/stacker/imports'. See https://github.com/project-stacker/stacker/issues/571 for migration.
'import' directive used in layer 'base' inside file '/home/serge/delme/stacker-imports/stacker-req.yaml' is deprecated. Support for 'import' will be removed in releases after 2025-01-01. Migrate by changing 'import' to 'imports' and '/stacker' to '/stacker/imports'. See https://github.com/project-stacker/stacker/issues/571 for migration.
'import' directive used in layer 'base' inside file '/home/serge/delme/stacker-imports/stacker-req.yaml' is deprecated. Support for 'import' will be removed in releases after 2025-01-01. Migrate by changing 'import' to 'imports' and '/stacker' to '/stacker/imports'. See https://github.com/project-stacker/stacker/issues/571 for migration.
preparing image base...
copying /home/serge/delme/stacker-imports/helloworld
loading docker://busybox:latest
Copying blob 9ad63333ebc9 done  
Copying config 3f57d9401f done  
Writing manifest to image destination
Storing signatures
+ cp /stacker/helloworld /
'import' directive used in layer 'buildenv' inside file '/home/serge/delme/stacker-imports/stacker.yaml' is deprecated. Support for 'import' will be removed in releases after 2025-01-01. Migrate by changing 'import' to 'imports' and '/stacker' to '/stacker/imports'. See https://github.com/project-stacker/stacker/issues/571 for migration.
preparing image buildenv...
+ cp /stacker/helloworld /h3
cache miss because layer definition was changed
error: couldn't find a cache of base layer for buildenv: base
error: exit status 1
```

### Expected behavior

When the two stackerfiles are combined into one, the build succeeds:
```
serge@serge-l-PF3DENS3 ~/delme/stacker-imports$ cat stacker-import-allinone.yaml 
base:
    from:
      type: docker
      url: docker://busybox:latest
    build_only: true
    import:
        - helloworld
    run: |
      cp /stacker/helloworld /

buildenv:
    from:
        type: built
        tag: base
    import:
        - stacker://base/helloworld
    run: |
        cp /stacker/helloworld /h3

serge@serge-l-PF3DENS3 ~/delme/stacker-imports$ ./stacker build -f !$
./stacker build -f stacker-import-allinone.yaml
'import' directive used in layer 'base' inside file 'stacker-import-allinone.yaml' is deprecated. Support for 'import' will be removed in releases after 2025-01-01. Migrate by changing 'import' to 'imports' and '/stacker' to '/stacker/imports'. See https://github.com/project-stacker/stacker/issues/571 for migration.
'import' directive used in layer 'buildenv' inside file 'stacker-import-allinone.yaml' is deprecated. Support for 'import' will be removed in releases after 2025-01-01. Migrate by changing 'import' to 'imports' and '/stacker' to '/stacker/imports'. See https://github.com/project-stacker/stacker/issues/571 for migration.
'import' directive used in layer 'base' inside file '/home/serge/delme/stacker-imports/stacker-import-allinone.yaml' is deprecated. Support for 'import' will be removed in releases after 2025-01-01. Migrate by changing 'import' to 'imports' and '/stacker' to '/stacker/imports'. See https://github.com/project-stacker/stacker/issues/571 for migration.
'import' directive used in layer 'buildenv' inside file '/home/serge/delme/stacker-imports/stacker-import-allinone.yaml' is deprecated. Support for 'import' will be removed in releases after 2025-01-01. Migrate by changing 'import' to 'imports' and '/stacker' to '/stacker/imports'. See https://github.com/project-stacker/stacker/issues/571 for migration.
preparing image base...
using cached copy of /home/serge/delme/stacker-imports/helloworld
loading docker://busybox:latest
Copying blob 9ad63333ebc9 skipped: already exists  
Copying config 3f57d9401f done  
Writing manifest to image destination
Storing signatures
cache miss because layer definition was changed
+ cp /stacker/helloworld /
preparing image buildenv...
failed to send signal urgent I/O condition no such process
cache miss because layer definition was changed
+ cp /stacker/helloworld /h3
filesystem buildenv built successfully
```

### Screenshots

_No response_

### Additional context

_No response_

mikemccracken:
Fixes #592 by serializing the `wasLegacyImport` field to ensure that the cache records the correct value when an image has it set.

Includes a test based on Serge's description in #592.

The symptom is an error "cache miss because layer definition was changed" when the definition could not have changed since build. This happens during multi file builds in builds with dependencies, and in build-then-publish flows, where the build will be fine but the publish will fail with the same error.

Using `--debug` to show the cache mismatch shows that somewhere along the line, the layer is being serialized with the wrong default value when writing to the cache.

Needs to change currentCacheVersion to ensure compatibility, see the TestCacheEntryChanged function comment for more info.

### DIFF
--- a/pkg/stacker/build.go
+++ b/pkg/stacker/build.go
@@ -116,7 +116,7 @@ func (b *Builder) updateOCIConfigForOutput(sf *types.Stackerfile, s types.Storag
 	}
 
 	inDir := types.InternalStackerDir
-	if l.WasLegacyImport() {
+	if l.WasLegacyImport {
 		inDir = types.LegacyInternalStackerDir
 	}
 
@@ -370,7 +370,9 @@ func (b *Builder) build(s types.Storage, file string) error {
 
 		log.Infof("preparing image %s...", name)
 		inDir := types.InternalStackerDir
-		if l.WasLegacyImport() {
+		if l.WasLegacyImport {
+			log.Debugf("image %s uses legacy import syntax, will also mount imports at %s",
+				name, types.LegacyInternalStackerDir)
 			inDir = types.LegacyInternalStackerDir
 		}
 
--- a/pkg/stacker/cache.go
+++ b/pkg/stacker/cache.go
@@ -22,7 +22,7 @@ import (
 	"stackerbuild.io/stacker/pkg/types"
 )
 
-const currentCacheVersion = 13
+const currentCacheVersion = 14
 
 type ImportType int
 
--- a/pkg/stacker/cache_test.go
+++ b/pkg/stacker/cache_test.go
@@ -125,5 +125,5 @@ func TestCacheEntryChanged(t *testing.T) {
 	// This test works because the type information is included in the
 	// hashstructure hash above, so using a zero valued CacheEntry is
 	// enough to capture changes in types.
-	assert.Equal(uint64(0x4ef432528243e356), h)
+	assert.Equal(uint64(0xa26696f335211127), h)
 }
--- a/pkg/types/layer.go
+++ b/pkg/types/layer.go
@@ -260,7 +260,7 @@ type Layer struct {
 	OS              *string           `yaml:"os" json:"os,omitempty"`
 	Arch            *string           `yaml:"arch" json:"arch,omitempty"`
 	Bom             *Bom              `yaml:"bom" json:"bom,omitempty"`
-	wasLegacyImport bool
+	WasLegacyImport bool              `yaml:"was_legacy_import" json:"was_legacy_import,omitempty"`
 }
 
 func parseLayers(referenceDirectory string, lms yaml.MapSlice, requireHash bool) (map[string]Layer, error) {
@@ -369,7 +369,7 @@ func parseLayers(referenceDirectory string, lms yaml.MapSlice, requireHash bool)
 		if len(layer.LegacyImport) != 0 {
 			layer.Imports = layer.LegacyImport
 			layer.LegacyImport = nil
-			layer.wasLegacyImport = true
+			layer.WasLegacyImport = true
 		}
 
 		ret[name], err = layer.absolutify(referenceDirectory)
@@ -440,11 +440,6 @@ func (l Layer) absolutify(referenceDirectory string) (Layer, error) {
 	}
 
 	return ret, nil
-}
-
-// WasLegacyImport - return true if this layer used legacy 'import' directive.
-func (l Layer) WasLegacyImport() bool {
-	return l.wasLegacyImport
 }
 
 func requireImportHash(imports Imports) error {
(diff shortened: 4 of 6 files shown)


hallyn (approved):

rchincha (approved):
lgtm

mikemccracken:
Some clarification, in part because I wasn't sure until after this was merged, why the field needed to be *serialized* to fix the build case - it doesn't:

There are two ways this ends up failing: for the single multi file run of `stacker build`, the `cache` package makes a _copy_ of the `Layer` struct when the build checks the cache during the second file's build - **copying the `Layer` out of the cache results in all non-exported fields being set to the zero value**, so `wasLegacy` ends up as `false`, which conflicts with its original value, which is what it is being compared with in the second stage. 

To fix this we only need to make it an exported field.

In the `build` then `publish` case, it is going to read in from the cache , and set the field to the zero value because it isn't serialized. To fix that, we also need to add the serialization annotation to write the field out to the cache.

It might be worth having a publish test case that also exercises this- I'll look into it.

mikemccracken:
My clarification was wrong yesterday. The behavior of copying in go structs with non-exported values is not quite that crazy - there are still some sharp edges there but if you just assign a struct, you will get the identical contents.

My confusion was that I didn't realize that stacker re-read the cache from disk in between files of multi-file builds. So in fact adding the serialization annotation itself is the fix in both cases - and we have to make the field exported to serialize it. So the fix here is correct and minimal, and as a bonus now I can explain it correctly.

PR #604, opened 2024-03-20
Hallyn 2024 03 18 test rmdir

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -285,6 +285,6 @@ require (
 )
 
 replace (
-	github.com/opencontainers/umoci => github.com/project-stacker/umoci v0.0.0-20240223003603-555a0af426e1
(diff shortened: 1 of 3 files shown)


hallyn:
Thanks for fixing my testcase - I'm sorry, I hadn't noticed that I had uploaded an older copy of my tree.  I had actually just switched to using docker://busybox:latest, but what you have here is great.

So as we discussed in the debug session - I think overlay is working just fine (we should verify this by also doing the test with --layer-type=overlay in the testcase).  I think the tar file is breaking because overlay is not creating /.wh_ccc when it has /ccc with contents and overlay.opaque=y.  When we only rm -rf /bbb, then it does create /.wh_bbb.

hallyn (approved):

Issue #617, opened 2024-05-08
Bug: max read sized reached when importing layer with very large files ss


raharper:
### stacker version

stacker version stacker v1.0.0-rc13-86ba851 liblxc cb8e38aca27a23964941f0f011a8919aab8bebab

### Describe the bug

When a stacker layer has a file > 10G, the following stack trace shows:

```
error: max read size reached
error: exit status 1
stackerbuild.io/stacker/pkg/container.MaybeRunInNamespace
        /stacker-tree/pkg/container/userns.go:102
main.main.func3
        /stacker-tree/cmd/stacker/main.go:324
github.com/urfave/cli/v2.(*Command).Run
        /stacker-tree/.build/gopath/pkg/mod/github.com/urfave/cli/v2@v2.25.0/command.go:213
github.com/urfave/cli/v2.(*App).RunContext
        /stacker-tree/.build/gopath/pkg/mod/github.com/urfave/cli/v2@v2.25.0/app.go:332
github.com/urfave/cli/v2.(*App).Run
        /stacker-tree/.build/gopath/pkg/mod/github.com/urfave/cli/v2@v2.25.0/app.go:309
main.main
        /stacker-tree/cmd/stacker/main.go:329
runtime.main
        /usr/lib/go/src/runtime/proc.go:267
runtime.goexit
        /usr/lib/go/src/runtime/asm_amd64.s:1650
```

I eventually tracked this down to an error from github.com/udhos/equalfile  

https://github.com/udhos/equalfile/blob/90a7461a2a226fa62eca6e350b5207ae9c78fff8/equalfile.go#L390

The default max read size in equalfile is 10^10

https://github.com/udhos/equalfile/blob/90a7461a2a226fa62eca6e350b5207ae9c78fff8/equalfile.go#L12

It looks like there are some equalfile Options struct which can use a "Limited Reader" to avoid the max size limit.

### To reproduce

1. Configuration

2. Client tool used
3. Seen error

```
error: max read size reached
error: exit status 1
stackerbuild.io/stacker/pkg/container.MaybeRunInNamespace
        /stacker-tree/pkg/container/userns.go:102
main.main.func3
        /stacker-tree/cmd/stacker/main.go:324
github.com/urfave/cli/v2.(*Command).Run
        /stacker-tree/.build/gopath/pkg/mod/github.com/urfave/cli/v2@v2.25.0/command.go:213
github.com/urfave/cli/v2.(*App).RunContext
[...]

PR #618, opened 2024-05-08
fix: avoid max read size reached error when cmp imported files

raharper:
**What type of PR is this?**

bug

#617 

when dealing with large files in imports (> 10**10 bytes) then equalfile fails.
this PR uses io.LimitedReader to avoid the error path.

**Automation added to e2e**:

No

No

```release-note

```

### DIFF
--- a/pkg/stacker/import.go
+++ b/pkg/stacker/import.go
@@ -1,6 +1,7 @@
 package stacker
 
 import (
+	"io"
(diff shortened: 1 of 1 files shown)


mikemccracken (approved):
nice catch, thanks!

rchincha (approved):
lgtm

PR #623, opened 2024-06-03
fix: update docs for bindsdoc

mikemccracken:
we were missing the dict version in the docs

**What type of PR is this?**

documentation

binds dict style was undocumented


mbshields (approved):
lgtm

PR #624, opened 2024-06-04
fix: check bind source file/dir exists

raharper:
Stacker does not check if a bind mount source exists and if the source is a directory then the container crashes with a stack trace that does not indicate that the missing source dir is the issue.

Check if the source exists and return an error indicating the source is missing instead of the container stack trace.

**What type of PR is this?**

bug

```
$ mkdir test-binds && cd test-binds
$ cat > stacker.yaml <<EOF
hello-binds:
  from:
    type: docker
    url: docker://zothub.io/tools/busybox:stable
  binds:
    - mydir -> /mydir
  run: |
    mkdir -p /hello-stacker-app
    echo 'echo "Hello Stacker!"' > /hello-stacker-app/hello.sh
EOF
$ stacker build -f stacker.yaml
```

Above test returns an error indicating that the source directory in binds is missing.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/stacker/build.go
+++ b/pkg/stacker/build.go
@@ -820,9 +820,10 @@ func SetupLayerConfig(config types.StackerConfig, c *container.Container, l type
 	}
 
 	for _, bind := range l.Binds {
+		log.Debugf("bind mounting %q into container at %q", bind.Source, bind.Dest)
 		err = c.BindMount(bind.Source, bind.Dest, "")
 		if err != nil {
-			return err
+			return errors.Errorf("failed to bind mount %q at %q: %s", bind.Source, bind.Dest, err)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/stacker/build.go
```diff
@@ -820,6 +820,12 @@ func SetupLayerConfig(config types.StackerConfig, c *container.Container, l type
 	}
 
 	for _, bind := range l.Binds {
+		_, err := os.Stat(bind.Source)
+		// c.BindMount allows files to not exist and will create them, but not so for directories.
```
mikemccracken, line 824: is this true? BindMount() sets create=dir for dirs and create=file for files, but that's just for the destination right? If the source isn't there it doesn't matter what type it is, I thought.

anyway, the code is right otherwise. lgtm

raharper: I can drop the comment; but it shows the general confusion;  the createOpts don't say if they're for the source or the dest; but thinking about it for a second, it's always for dest.

I might touch those variables...

Alternatively, I can move the check into the BindMount() call, which would protect other paths as well.

Thoughts?

mikemccracken: yeah probably a good idea to just do the check in bindmount() as you suggest. not as likely to catch, but why not...

hallyn (approved):

PR #625, opened 2024-06-11
chore: add test for failing on missing bind source

mikemccracken:
Just copied Ryan's test from the description of #624 with minimal editing.


raharper (approved):
Nice! Thanks for adding the test I should have included.  
"thats_no_dir" is a nice touch!

PR #643, opened 2024-09-10
build: improve errors for --oci-dir cases

mikemccracken:
If the argument to oci-dir exists but is empty, we currently get a confusing error about it not being a layout. Other cases could be better explained while we're at it too.


rchincha (approved):
lgtm

PR #646, opened 2024-09-17
test: add additional tests for whiteout fixes

rchincha:
acd8a6856fb356805bd7eeba5c83efc519385642 address some breakages related to whiteouts.

This PR adds some more verification steps.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/test/whiteout.bats
+++ b/test/whiteout.bats
@@ -69,6 +69,16 @@ fulldir:
     rm -rf /a1
     mkdir /a1
     touch /a1/newfile
+
+redo:
+  from:
+    type: built
+    tag: bb
+  run: |
+    rm -rf /a1
+    mkdir /a1
+    touch /a1/newfile
+
 EOF
 
   stacker build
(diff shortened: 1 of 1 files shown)


test/whiteout.bats
```diff
@@ -69,19 +69,28 @@ fulldir:
...
+    touch /a1/newfile
+
 EOF
 
   stacker build
 
   # "bb" should not have a whiteout entry for /a1
   md=$(cat oci/index.json | jq .manifests[0].digest | sed s/sha256://g | tr -d \")
   ld=$(cat oci/blobs/sha256/"$md" | jq .layers[-1].digest | sed s/sha256://g | tr -d \")
-  "bsdtar -tvf oci/blobs/sha256/$ld | grep '.wh.a1'"
```
hallyn: Hi,

these look good.  Though for this particular hunk, would it be better to not remove this bsdtar|grep, but rather turn it into a bsdtar | grep -v '.wh.a1' ?

Issue #645, opened 2024-09-17
Feat: make lxc control socket unique per container


rchincha:
### Is your feature request related to a problem? Please describe.

If multiple stacker instances are invoked for the same tag/target, they may step on each other since the lxc control socket is not unique per container.

### Describe the solution you'd like

Make the lxc control socket unique per container.

mikemccracken:
note that the simplest way to do this is probably to make the container name used for a build include some randomness. 

lxc uses the lxcpath and container name to name the control socket, and nothing outside of stacker really cares what the container name is.

however for the lxcpath, stacker uses the `--roots-dir` argument - since this comes from the user, uniquifying it might be slightly more involved. (we could make a unique subdir, but we'd have to thread that through the stacker code)

How does it happen that two stackers use the same roots-dir argument but shouldn't conflict?
in some cases, concurrent runs of stacker on the same host might have the same path for the roots-dir arg, even though they aren't actually sharing it - e.g. if they are running in separate mount ns's.

PR #651, opened 2024-11-07
fix: randomize lxc container names

mikemccracken:
**What type of PR is this?**

bug

In situations where concurrent stacker runs are happening on a system, and they are building containers with the same name, and they are being done inside a mount namespace,
and the path name given as the roots dir is the same, but the actual mounted volume is different,
then both stackers will be able to acquire the file lock at $rootdir/.lock, and will go ahead and start containers named $name, which will then race to set up the lxc control socket, which is named after the container name and the rootfs path, which are both the same here.

The fix is to add some randomness to the lxc container name, which ensures that the socket won't clash. This should not affect other uses of the image name, which will still use the un-randomized name.

#645 

manual only, CI pending


mikemccracken:
Draft because it might need work if it breaks CI. the test suite is not reliable on my system right now so I am just using the github runners

mikemccracken:
The CI failures are real, there is a case now where a build creates a new layer when it shouldn't. empty-layer.bats checks for this and that's what's failing.

However it isn't obvious to me what exactly is going wrong, I spent a half day looking at it before Thanksgiving and didn't conclude. Will revisit later.

Maybe the Right Way is to have lxc include the mount ns id in its socket naming code?

PR #654, opened 2024-12-19
move to atomfs with fixed umount, update golang to 1.22

mikemccracken:
see https://github.com/project-machine/atomfs/pull/27

The MetadataPath field in the MountOCIOpts struct is not needed anymore, if left empty it will use /run/atomfs/ for all mounts.

The atomfs.bats file is not passing locally even though the equivalent test does pass in atomfs#27. Creating this PR to see if it passes in jenkins. It is something about the base image's verity device being in use. I suspect another test is not unmounting the shared base busybox image - maybe we need to isolate the tests better here, or just delete the atomfs test in stacker because it's covered in atomfs now


mikemccracken:
Converted to draft because we shouldn't merge this, we should wait until atomfs merges atomfs#27 and then update to the main repo

mikemccracken:
this should work now but will need a change to the first commit to bring in the real atomfs repo once https://github.com/project-machine/atomfs/pull/27 is merged.

go.mod
```diff
@@ -260,24 +261,24 @@ require (
...
-	golang.org/x/crypto v0.18.0 // indirect
+	golang.org/x/crypto v0.31.0 // indirect
 	golang.org/x/exp v0.0.0-20231206192017-f3f8817b8deb // indirect
-	golang.org/x/mod v0.14.0 // indirect
-	golang.org/x/net v0.20.0 // indirect
-	golang.org/x/sync v0.6.0 // indirect
-	golang.org/x/text v0.14.0 // indirect
-	golang.org/x/tools v0.17.0 // indirect
+	golang.org/x/mod v0.17.0 // indirect
+	golang.org/x/net v0.25.0 // indirect
```
raharper, line 267: can we go ahead a bump to x/net to v0.33.0

go.mod
```diff
@@ -1,8 +1,8 @@
 module stackerbuild.io/stacker
 
-go 1.21.0
+go 1.22
 
-toolchain go1.21.6
+toolchain go1.22.6
```
raharper, line 5: I think you need 1.22.8 for the x/net@v0.33.0 to fix the net cve

test/atomfs.bats
```diff
@@ -103,7 +103,7 @@ EOF
     # first layer should still exist since a is still mounted
     manifest=$(cat oci/index.json | jq -r .manifests[0].digest | cut -f2 -d:)
     first_layer_hash=$(cat oci/blobs/sha256/$manifest | jq -r .layers[0].digest | cut -f2 -d:)
-    [ ! -b "/dev/mapper/$last_layer_hash-verity" ]
+    [ -b "/dev/mapper/$first_layer_hash-verity" ]
```
raharper, line 106: could you comment on what this fix is doing?

test/bom.bats
```diff
@@ -258,14 +258,18 @@ EOF
...
+    if [ -n "${ZOT_HOST}${ZOT_PORT}" ]; then
       stacker publish --skip-tls --url docker://${ZOT_HOST}:${ZOT_PORT} --tag latest
       refs=$(regctl artifact tree ${ZOT_HOST}:${ZOT_PORT}/parent:latest --format "{{json .}}" | jq '.referrer | length')
       [ $refs -eq 2 ]
       refs=$(regctl artifact get --subject ${ZOT_HOST}:${ZOT_PORT}/parent:latest --filter-artifact-type "application/spdx+json" | jq '.SPDXID')
       [ $refs == \"SPDXRef-DOCUMENT\" ]
     fi
 
+    if [ -z "${ZOT_HOST}${ZOT_PORT}" ]; then
+    skip "second half of test requires running zot"
```
raharper, line 270: missing a tab? whitespace issue?

test/bom.bats
```diff
@@ -1,13 +1,13 @@
 load helpers
 
 function setup_file() {
-  if [ -n "${ZOT_HOST}:${ZOT_PORT}" ]; then
+  if [ -n "${ZOT_HOST}${ZOT_PORT}" ]; then
```
raharper, line 4: what's the reason for removing the colon? it seems strange to have both host and port in the string to check.

raharper (changes requested):
Looks great!  Thank you for working through the CI.  Only change I think needed is ensuring we get the x/net bump to v0.33.0; I think you need 1.22.8 toolchain.  The rest are some questions.

go.mod
mikemccracken: ah ok, I will update my tools and redo the go mod tidy, thanks

test/atomfs.bats
mikemccracken: sure - so, per the comment, the first layer should still exist since `a` is still mounted - line 104 and 105 put the hash of the first layer into `first_layer_hash`, and we should be checking ` [ -b "/dev/mapper/$first_layer_hash-verity" ]` to test that the verity block device for `first_layer_hash` still exists.

the earlier code was instead using an unassigned variable `$last_layer_hash`, which is set later in this test and in other tests, but at this point will be empty - so along with the `!`, we ended up checking that the block device `/dev/mapper/-verity` does NOT exist, which will be true, so the test passes but didn't test what it should. I assume this was just a typo.

this is the reason for the initial run of shellcheck I added, that specifically checks for the case where a variable is defined but in a separate subshell, which is what this turns into once you evaluate the bats magic:
https://github.com/project-stacker/stacker/pull/654/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R192

test/bom.bats
mikemccracken: shellcheck pointed out that using `-n` with a string that includes a literal string will never be false, because `[ -n ":" ]` is still true - so including the colon meant that we always try to do the zot stuff, even if there's no zot host/port configured.

I guess we could also just remove the port and only check `-n "$ZOT_HOST"`, if we wanted to support using a zot without an explicit port. This just seemed like the simplest change that fixed the issue

test/bom.bats
mikemccracken: hrm. looks like emacs shell-script mode's auto indent gets confused in bats files

go.mod
mikemccracken: done

test/bom.bats
raharper: sounds reasonable.  Thanks

raharper (approved):

rchincha (approved):
lgtm

PR #694, opened 2025-05-20
feat: update to github.com/opencontainers/umoci@v0.5.0

cyphar:
This allows us to switch away from our umoci fork now that upstream
supports OverlayfsRootfs and the various features we need.

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>


hallyn:
Thanks! This looks great.

cyphar:
Updated to use the actual release. This is missing some of the patches from your fork, so I would suggest doing some extra testing to make sure that I didn't subtly break anything (in particular, all of the lowerdir tracking isn't in umoci because it wasn't necessary to implement the minimal version of plain-to-opaque whiteout conversion).

cyphar:
WDYT @rchincha?

chinchani:
https://github.com/project-stacker/stacker/commit/589a6486aeb2a5fc157edf043d0310795cd2c90b

CI Failure related to this. IIRC, we were trying to keep stacker and skopeo in sync wrt layer compression else they would change underbnath us.

rchincha:
@cyphar 

https://github.com/project-stacker/stacker/blob/main/Makefile#L45

Can this be bumped to 1.13.0?

rchincha (approved):
lgtm

rchincha:
The main changes afaict between skopeo 1.9.3 ... 1.13.0 are:
For 1.9.3: klauspost/pgzip@1.2.5 which calls into klauspost/compress@1.15.9
For 1.13.0: klauspost/pgzip@1.2.6 which calls into klauspost/compress@1.16.6

rchincha:
Have run the CI multiple times without any failures now.

pkg/overlay/pack.go
```diff
@@ -34,12 +34,6 @@ import (
 
 var tarEx sync.Mutex
 
-// Container image layers are often tar.gz, however there is nothing in the
-// spec or documentation which standardizes compression params which can cause
-// different layer hashes even for the same tar. So picking compression params
-// that most tooling appears to be using.
-const gzipBlockSize = mutate.GzipBlockSize(256 << 12)
```
raharper, line 41: What size is being used in umoci@v.0.5.0 now?

Any idea why dropping this and bumping skopeo tripped up the test-cases?

raharper: Found the issue with details https://github.com/opencontainers/umoci/pull/509

pkg/overlay/pack.go
```diff
@@ -412,34 +406,19 @@ func generateLayer(config types.StackerConfig, _ casext.Engine, mutators []*muta
...
-		blob, mediaType, rootHash, err := generateBlob(layerType, dir, config.OCIDir, lowerDirs)
+		blob, mediaType, rootHash, err := generateBlob(layerType, dir, config.OCIDir)
 		if err != nil {
 			return false, err
 		}
 		defer blob.Close()
 
 		if layerType.Type == "tar" {
-			desc, err = mutator.Add(context.Background(), mediaType, blob, history, mutate.GzipCompressor.WithOpt(gzipBlockSize), nil)
+			desc, err = mutator.Add(context.Background(), mediaType, blob, history, mutate.GzipCompressor, nil)
```
raharper, line 421: Might be useful to leave a breadcrumb pointing to (https://github.com/opencontainers/umoci/pull/509).
Something to give some background on why we got to drop custom umoci fork since umoci picked up the docker-compatible size we used.

rchincha: Either leave a comment in the code here or commit msg.

cyphar: https://github.com/opencontainers/umoci/pull/581 would be the more correct thing to point to, but yes I'll update the commit message.

PR #726, opened 2025-09-12
update use of auth.json credentials for `import`, add test of existing image import behavior

mikemccracken:
containers/image `copy.Image()` already uses `GetCredentials()` to look in `auth.json`  for base image imports.

this adds a test to cover this case using a zot configured to require auth. also generates certs for the zot in the test.

It also updates the previous auth code for `import` section imports, so that we now pass the entire path of an import to `GetCredentials()`, which lets us use different creds for different subpaths on a host.

### DIFF
--- a/doc/stacker_yaml.md
+++ b/doc/stacker_yaml.md
@@ -40,10 +40,12 @@ Some directives are irrelevant depending on the type. Supported types are:
 
 - `url` is required
 
-- `insecure` is optional. 
+- `insecure` is optional. If unspecified, the default is is `false`.
 
 When `insecure` is specified, stacker attempts to connect via http instead of
-https to the Docker Hub.
+https to the image registry in the `url`.
+
+When `insecure` is false, Stacker can authenticate to the image registry using HTTP Basic Authentication. See the section "Credential Handling" for further details.
 
 #### `type: tar`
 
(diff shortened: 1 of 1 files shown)


mikemccracken:
This is ready to review now.

test/setup_suite.bash
```diff
@@ -1,6 +1,58 @@
 #!/bin/bash
 
+function write_certs {
```
mikemccracken, line 3: this is borrowed verbatim from the zot test suite, fwiw

pkg/stacker/network.go
```diff
@@ -85,12 +86,13 @@ func Download(cacheDir string, remoteUrl string, progress bool, expectedHash, re
 	if err != nil {
 		return "", err
 	}
-
-	creds, err := config.GetCredentials(&types.SystemContext{}, u.Hostname())
+	key := fmt.Sprintf("%s%s", u.Host, u.Path)
```
mikemccracken, line 89: this is the main change of this PR - adding path to the key we pass to GetCredentials

Makefile
```diff
@@ -124,7 +124,7 @@ go-test:
 	go tool cover -html coverage.txt  -o $(HACK_D)/coverage.html
 
 .PHONY: download-tools
-download-tools: $(GOLANGCI_LINT) $(REGCLIENT) $(ZOT) $(BATS) $(UMOCI)
+download-tools: $(GOLANGCI_LINT) $(REGCLIENT) $(ZOT) $(BATS) $(UMOCI) $(SKOPEO)
```
mikemccracken, line 127: we had code to build skopeo but were not triggering it during regular `make test`

rchincha (approved):
lgtm

PR #729, opened 2025-10-07
fix: build-deps for 24.04 don't need ppa anymore

raharper:
It looks like newer lxc release in Noble no longer needs the lxc ppa; it can build stacker-dynamic/stacker static without the package which currently conflicts with the Noble lxc package

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/install-build-deps.sh
+++ b/install-build-deps.sh
@@ -53,11 +53,6 @@ installdeps_ubuntu() {
             sudo add-apt-repository -y ppa:project-machine/squashfuse
             ;;
         24.04)
-            # lp:2080069
-            # temporarily add puzzleos/dev to pickup lxc-dev package which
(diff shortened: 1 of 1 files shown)


hallyn (approved):
Thanks!

raharper:
> ```
> /usr/bin/ld: cannot find -llxc: No such file or directory
> ```
> 
> What the heck?

We still need liblxc.a -- that was what our PPA did for 24.04 (and 22.04) -- as we discussed, I'm seeing of 25.10 lxc has the fix we pushed here:

https://bugs.launchpad.net/ubuntu/+source/lxc/+bug/2080069

hallyn:
I'm ok with this, but I think it needs to be well documented for people who want to build it what's going on.

PR #762, opened 2026-03-23
fix: inline convert test Dockerfile data

raharper:
Skip downloading Dockerfiles for the convert test as they add instability to the tests when github is unreliable.  Capture what we are fetching and the sha256sum values.

**What type of PR is this?**

bug

Help removing instability in the convert test case

We had a recent nightly CI failure due to github repo instability so instead of downloading (git clone, etc) capture the content in our repository.

https://github.com/project-stacker/stacker/actions/runs/23416513187/job/68112948797

```
# + sh -e -c curl --retry 10 -S -L --output /tmp/elasticsearch.tar.gz https://artifacts-no-kpi.elastic.co/downloads/elasticsearch/elasticsearch-8.17.10-linux-$(arch).tar.gz
#   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
#                                  Dload  Upload   Total   Spent    Left  Speed
# 
...
 84  462M   84  389M    0     0  13.4M      0  0:00:34  0:00:29  0:00:05 13.1M
# curl: (92) HTTP/2 stream 1 was not closed cleanly: INTERNAL_ERROR (err 2)
```

CI

**Automation added to e2e**:

None

No

No

```release-note

```

### DIFF
--- a/test/convert.bats
+++ b/test/convert.bats
@@ -79,7 +79,12 @@ EOF
   skip_slow_test
   # https://github.com/elastic/dockerfiles/archive/refs/tags/v8.17.10.tar.gz
   # SHA256SUM=75f89195b53c9d02d6a45d2bc9d51a11d92b079dcc046893615b04455269ba1a  dockerfiles-8.17.10/elasticsearch/Dockerfile
-  DOCKERFILE="data/elasticsearch-8.17.10.Dockerfile"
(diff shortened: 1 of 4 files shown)


raharper:
Looking inside the dockerfiles, this won't help since they fetch the same resources from github;  they already have lots of retries...

PR #764, opened 2026-05-14
Fix/bump docker and container storage

raharper:
**What type of PR is this?**

cleanup

CVEs

#742 
#734 
#732 
#708 
#704 

Address upstream go CVEs

security scan on stacker binary

make test && make lint

**Automation added to e2e**:

none

no

no

```release-note

```


raharper:
btw, `make test` passes locally on an Ubuntu 24.04 Cloud VM... so looks like we're rolling dice with Azure vms...

raharper:
> btw, `make test` passes locally on an Ubuntu 24.04 Cloud VM... so looks like we're rolling dice with Azure vms...

Not clear what's going on in the VM here, but something was up with how it was using git.  I've changed test to not use `run_git` anymore and pass in the VERSION_FULL which ran the same commands that test ran; used to confirm that stacker inserts it's git build VERSION value into the OCI annotations.

rchincha (approved):

PR #766, opened 2026-06-25
Fix/migrate to go podman image

raharper:
**What type of PR is this?**

cleanup

n/a

Migrate from containers/image/v5 (now deprecated) to the new mono-repo
https://github.com/podman-container-tools/container-libs

Bump incus/v6 -> incus/v7

Bump go 1.25.10 -> 1.25.11

Bump x/crypto and x/net to fix CVEs

build and make test

**Automation added to e2e**:

none

no

no

```release-note

```


raharper:
Chasing this odd error in the overlay test:

```
# Error: couldn't import base layer base: initializing source 
docker://gcr.io/distroless/base:latest: 
loading registries configuration: 
loading registries configuration "/etc/containers/registries.conf":
registries.conf must be in v2 format but is in v1
```

I cannot recreate this on a local 24.04 x86_64 vm, nor on a 24.04 arm64 vm -- currently seeing if the ubuntu 24.04 runner has some of these things baked into the image that aren't in normal/vanilla 24.04 environments.

raharper:
> Chasing this odd error in the overlay test:
> 
> ```
> # Error: couldn't import base layer base: initializing source 
> docker://gcr.io/distroless/base:latest: 
> loading registries configuration: 
> loading registries configuration "/etc/containers/registries.conf":
> registries.conf must be in v2 format but is in v1
> ```
> 
> I cannot recreate this on a local 24.04 x86_64 vm, nor on a 24.04 arm64 vm -- currently seeing if the ubuntu 24.04 runner has some of these things baked into the image that aren't in normal/vanilla 24.04 environments.

Yeah, I can reproduce this by placing a v1 registries.conf in $HOME/.config/containers/registries.conf

```
[registries.search]
registries = ['docker.io', 'quay.io']
```

Testing out a change to overlay.bats to move it out of the way at start and restore at the end.

Not super happy about that w.r.t running tests locally could mess with the system.

raharper:
> > Chasing this odd error in the overlay test:
> > ```
> > # Error: couldn't import base layer base: initializing source 
> > docker://gcr.io/distroless/base:latest: 
> > loading registries configuration: 
> > loading registries configuration "/etc/containers/registries.conf":
> > registries.conf must be in v2 format but is in v1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I cannot recreate this on a local 24.04 x86_64 vm, nor on a 24.04 arm64 vm -- currently seeing if the ubuntu 24.04 runner has some of these things baked into the image that aren't in normal/vanilla 24.04 environments.
> 
> Yeah, I can reproduce this by placing a v1 registries.conf in $HOME/.config/containers/registries.conf
> 
> ```
> [registries.search]
> registries = ['docker.io', 'quay.io']
> ```
> 
> Testing out a change to overlay.bats to move it out of the way at start and restore at the end.
> 
> Not super happy about that w.r.t running tests locally could mess with the system.

Current "fix" here is for github runners to move it out of the way.  

stacker using podman containers lib means that we should consume those configs if present; just like we do for things like containers auth.json;

Not sure if we should extend `stacker check`  or just let it fail like it did and user can sort out updating their registries.conf file.

hallyn (approved):

